PR #18, opened 2015-04-30
Add validation prototype (fixes #13)

### LINKED ISSUE #13: Implement card validation lib.
Implement the braintree card validation lib [1] matching it up to our look and feel. 

[1] https://github.com/braintree/card-validator

muffinresearch:
This is the beginnings of the validation.

I've pushed it up as built docs here [1] so you can see it and try it out (e.g. try adding a card with a number like 52XX)

Note it doesn't enable the submit yet, and more x-browser testing required but worth checking out the code as it's written nethertheless. Oh yeah and tests to follow naturally.

[1]  http://mozilla.github.io/payments-ui/forms.html


muffinresearch:
Just pushed a couple of fixes to make sure card icons are correctly shown when switching between card examples as detailed here: https://developers.braintreepayments.com/javascript+python/guides/credit-cards#no-credit-card-errors

public/img/credit-cards-sprite.svg
```diff
@@ -0,0 +1,102 @@
...
+    <rect id="Rectangle" height="8" width="36" y="19" x="8" fill="#fff"/>
+    <rect id="Rectangle" height="2" width="29" y="22" x="11" fill="#424143"/>
+    <path id="Shape" fill="#DFDEDD" d="m54.75 1.209c2.206 0 4 1.794 4 4v30c0 2.206-1.794 4-4 4h-50c-2.206 0-4-1.794-4-4v-30c0-2.206 1.794-4 4-4h50zm0-1h-50c-2.75 0-5 2.25-5 5v30c0 2.75 2.25 5 5 5h50c2.75 0 5-2.25 5-5v-30c0-2.75-2.25-5-5-5z"/>
+    <rect id="Rectangle" height="8" width="8" y="19" x="47" fill="#fff"/>
+    <rect id="Rectangle" height="2" width="6" y="22" x="48" fill="#424143"/>
+    <circle id="Oval" stroke-width="2" cy="22.75" stroke="#E9564B" cx="49.75" r="7.75"/>
+   </g>
+  </g>
+ </g>
+</svg>
```
andymckay: license on this stuff ok?

public/scss/_cardicons.scss
```diff
@@ -0,0 +1,46 @@
+
```
andymckay: extra blank line

public/scss/_tooltip.scss
```diff
@@ -0,0 +1,60 @@
...
+      text-decoration: underline;
+      &.dismiss {
+        margin-left: 5px;
+        text-decoration: none;
+      }
+    }
+  }
+
+
+  @include respond-to('trustedUI') {
```
kumar303: mmm, you had to sneak in trustedUI to our shiny new world of web payments? :cactus: Could it be called mobileWindow or something?

kumar303:
r+

Looks like a great start. I think if we choose to go the React route then we would probably do the validation differently. I think you do something more like declare a template based on data states (valid, maybe valid...) and it will update accordingly based on the card number that gets inputted to the component -- that is, no explicit class toggling with jquery. We could always change it later.

public/scss/_tooltip.scss
muffinresearch: Ah that's just a file copied verbatim from fxa-content server. Not yet used.

PR #26, opened 2015-05-06
Add masking (fixes #19)

### LINKED ISSUE #19: Add input masking to form fields.
The card-validator lib is handy but we should look to add some input masking to make it really obvious how to fill in the forms.

muffinresearch:
Whilst looking at the braintree drop-in code to get some ideas for how they are handling pasted input (and the other gazillion edge cases that doing input masking brings) and it turned out they're using formatter.js [1]. So to cut a long story short I've switched to that in order to keep as much parity with the drop-in UI as possible and to avoid adding more code for us to maintain.

This branch:

* Adds full input masking for the card, expiry and cvv/cid
* Adds card change events so that the mask pattern gets updated for cards that feature different patterns and updates the cvv/cid placeholder.

[1] https://github.com/firstopinion/formatter.js


muffinresearch:
You can see the code in action here: http://mozilla.github.io/payments-ui/forms.html

andymckay:
r+, but does this need some tests?

muffinresearch:
> r+, but does this need some tests?

Yeah it definitely does. They'll come as soon as I have something setup to run them.

PR #29, opened 2015-05-07
add in a data only container

andymckay:


andymckay:
For https://github.com/mozilla/payments-env/pull/10

andymckay:
For https://github.com/mozilla/payments-env/pull/10

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM mozillamarketplace/centos-python27-mkt:0.5
+
+COPY . /srv/payments-ui
```
muffinresearch: Should be the contents of public not everything from the project root.

Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM mozillamarketplace/centos-python27-mkt:0.5
```
muffinresearch: Iis there any advantage to building on top of the centos image as opposed to busybox? I guess all these layers (for centos/python) are there (cos we're using them elsewhere) - so size shouldn't be any issue.

Dockerfile
andymckay: Ah right.

Dockerfile
andymckay: As we chatted its cached. We've already got busybox which is tiny, may as well use that.

muffinresearch:
lgtm r+

PR #34, opened 2015-05-14
Move to react (fixes #30, #27)

### LINKED ISSUE #30: Bring react interface into payments UI

muffinresearch:
r? @kumar303 

Bit of a big branch this but it comprises: 

* Moving to React (this will break things around the styleguide but I'll come back to that)
* Using webpack as a module loader and development process.
* Adding and enabling a submit button based on the form being valid.
* Removing the lib files that were copied into place. (As I mentioned, I'm thinking committing the static file modules to the tree is probably going to be the best plan to fix this rather than doing more copying) For example webpack now consumes the card-validator module from src and pulls in individual deps. This also means if we were to use the same deps as card-validator (it uses lodash) we wouldn't end up duplicating the libs.


.gitignore
```diff
@@ -3,4 +3,6 @@ bower_components
 .grunt
 styleguide/build
 public/css/*
+public/js/components/*.js
```
kumar303: Do you need to commit a README or something in the ``components`` directory so that it gets created on repo clones?

config/clean.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  options: {
+    //'no-write': true,
```
kumar303: is this still needed?

public/index.html
```diff
@@ -0,0 +1,16 @@
...
+    <meta name="viewport" content="width=device-width, user-scalable=no, initial-scale=1, maximum-scale=1">
+    <meta name="apple-mobile-web-app-capable" content="yes">
+    <link rel="stylesheet" href="css/main.css">
+    <title>Payments UI</title>
+  </head>
+  <body data-static-url='.'>
+    <main>
+      <div id='view'></div>
+    </main>
+    <script src="js/bundle.js"></script>
```
kumar303: how does this work for development? Will we see tracebacks to the original source files?

public/jsx/card-details.jsx
```diff
@@ -0,0 +1,98 @@
...
+              <MaskedInput type={type} className={fieldClass}
+                           id={field.id} pattern={field.pattern} placeholder={field.placeholder}
+                           onChange={that.handleChange.bind(that, index)} />
+            ) : (
+              <input type={type} className={fieldClass} id={field.id}
+                     placeholder={field.placeholder}
+                     onChange={that.handleChange.bind(that, index)} />
+            )
+          }
+        </label>
```
kumar303: this is pretty nice to read. I like it.

kumar303:
I'm still iffy about committing dependency source to the tree but if it's the only way to lock down versions then I guess we have to.

r+wc

public/index.html
muffinresearch: Yes it does, this is the point you're always using a bundle rather than dynamically requesting individual modules. 

I need to look into sourcemaps for the bundle though. Logged line numbers seem to break currently with the webpack-dev-server.

PR #38, opened 2015-05-15
Hook form up to braintree (fixes #35)

### LINKED ISSUE #35: Get a client-token from api and serialize card info

muffinresearch:
This is all quite lightweight but uses jquery and braintree-web to:
* Show a spinner as everything loads.
* Get the client token
* Show the form
* Tokenise the form with braintree 
* Get ready to send this on to our service.

We'll clean this up and modularize it all a bit more as we flesh out the rest of the infrastructure etc.


public/js/main.js
```diff
@@ -27,5 +29,34 @@ var fields = [
...
+    'method': 'post',
+    // TODO update this with a real endpoint.
+    'action': '/braintree/',
+  });
+
+  React.render(root, document.getElementById('view'));
+
+  $('#braintree-form').on('submit', function(e) {
+    e.preventDefault();
+    braintree.setup($(this).data('token'), 'custom', {
```
kumar303: is there a more react-like way to get this data rather than parsing the dom? re: ``$(this).data('token')``

kumar303:
looks like a good start.

r+wc

By the time you read this Monday you won't have the form submit endpoint yet. I got a little stuck on making logins (sessions) work. One thing that would be cool is to put the UI on the same domain as the service because once I land the login we'll need to rely on same-domain session cookies. 

Also, you may want to start looking at the login before looking at form processing: http://payments-service.readthedocs.org/en/latest/topics/api.html#sign-in

I think we can stand this up quickly by changing [the example code](https://github.com/mozilla/payments-example/blob/master/public/js/index.js#L43) to redirect to http://pay.dev/?token={token} which should be the UI (React) HTML page. That page would start by grabbing the token from GET and posting to ``/auth/sign-in`` so that we get the session cookie. The token generator call will fail until the user is logged in but that's the patch I'm still working on.

public/js/main.js
muffinresearch: Yeah, I'd imagine all the form submit hooks will move inside react completely at some point.

muffinresearch:
> By the time you read this Monday you won't have the form submit endpoint yet. I got a little stuck on making logins (sessions) work. One thing that would be cool is to put the UI on the same domain as the service because once I land the login we'll need to rely on same-domain session cookies. 

That's already the case. [1] But I can take on getting the urls moved around.

https://github.com/mozilla/payments-env/blob/master/docker/nginx/nginx.conf#L20

PR #51, opened 2015-05-21
Authenticate (fixes #50)

### LINKED ISSUE #50: Hook-up to auth

muffinresearch:
This is super-noddy just to get us back to having a working form.

I'm Going to look at routing solutions so we can start hanging this together with more structure (might make query-string lib redundant)

### DIFF
--- a/node_modules/query-string/index.js
+++ b/node_modules/query-string/index.js
@@ -1,0 +1,48 @@
+'use strict';
+
+exports.parse = function (str) {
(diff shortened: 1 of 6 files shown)


kumar303:
This looks great to get things working. It will need error handling for a few cases so it might be worth filing issues after merging:

* show error when access_token is missing from query string
* show error when login fails on the server

kumar303:
r+wc

PR #54, opened 2015-05-21
Add react-router routing (fixes #45 #52 #53)

### LINKED ISSUE #45: Use build/dist dir for built css/js

muffinresearch:
r? kumar 

This still needs error handling (and of course tests) but switches us over to using react-router and more of a view/component setup which I think starts to add shape to the app.

Requires the nginx config change here for urls to response correctly: https://github.com/mozilla/payments-env/pull/20


public/jsx/app.jsx
```diff
@@ -0,0 +1,35 @@
...
+  <Route handler={App}>
+    <Route name="login" path="/" handler={Login}/>
+    <Route name="card-form" path="/payment-card/" handler={CardDetails}/>
+  </Route>
+);
+
+module.exports = {
+  init: function() {
+    Router.run(routes, Router.HistoryLocation, function(Root) {
+      // Doesn't actually render anything.
```
kumar303: could you add a comment about why this exists if it doesn't render anything?

public/jsx/views/card-details.jsx
```diff
@@ -0,0 +1,56 @@
...
+
+  componentDidMount: function() {
+    $.ajax({
+      method: 'post',
+      url: this.props.apiSource,
+      context: this,
+    }).then(function(data) {
+      if (this.isMounted()) {
+        this.setState({'braintree_token': data.token}); // eslint-disable-line
+      }
```
kumar303: what would the else branch of this do? Hopefully not fail silently :)

public/jsx/views/card-details.jsx
```diff
@@ -0,0 +1,56 @@
...
+  },
+
+  componentDidMount: function() {
+    $.ajax({
+      method: 'post',
+      url: this.props.apiSource,
+      context: this,
+    }).then(function(data) {
+      if (this.isMounted()) {
+        this.setState({'braintree_token': data.token}); // eslint-disable-line
```
kumar303: I think this would look fine to satisfy the linter:
````
this.setState({
  'braintree_token': data.token,
});
````

File: public/jsx/views/card-details.jsx (hunk 2 above)
Comment (kumar303) [inline public/jsx/views/card-details.jsx]: maybe it shouldn't make the request at all if not mounted?

File: public/jsx/views/login.jsx [hunk 4]
```diff
@@ -0,0 +1,59 @@
...
+        access_token: router.getCurrentQuery().access_token,
+      },
+      method: 'post',
+      url: this.props.apiSource,
+      context: this,
+    }).then(function() {
+      if (this.isMounted()) {
+        this.setState({'is_logged_in': true}); // eslint-disable-line
+        this.transitionTo('card-form');
+      }
```
kumar303: again, it seems weird to me that you'd make a request and throw away the result

tests/test.card-form.jsx
```diff
@@ -1,7 +1,7 @@
 'use strict';
 
 
-var CardDetails = require('card-details');
+var CardDetails = require('components/card-form');
```
kumar303: maybe rename this var to ``CardForm`` ?

kumar303:
This looks pretty slick. I like the separate of concerns. The one-way data flow makes it read a lot better than the Backbone router too.

r+wc

The ``isMounted()`` bit was the only thing that looked odd to me since it seemed like in one case we'd make an API request and throw away the result.

I left https://github.com/mozilla/payments-ui/pull/55 open but you'll probably want to close it and rebuild the patch on top of this router patch.

public/jsx/views/card-details.jsx
muffinresearch: It's more that the best practice is not to change the state of something if it's unmounted. That would mean the component was destroyed.

public/jsx/views/card-details.jsx
muffinresearch: It's not that - setting state in componetDidMount is picked up by the react-eslint-plugin as this: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/no-did-mount-set-state.md

But in our case we're ok because thrashing won't happening + this is following the way the React docs say it should be done.

tests/test.card-form.jsx
muffinresearch: D'oh - yep, good catch.

public/jsx/views/card-details.jsx
muffinresearch: Thinking an explicit eslint comment would be better here as it would make it clearer why it was needed.

PR #57, opened 2015-05-22
Add initial extraction (fixes #47)

### LINKED ISSUE #47: Add in localisation framework
This is about adding a way to do localisations on the front end and the scripts for passing it up to verbatim so it can be localised. We should start this sooner rather than later so that we've got time to tell people about the localisation project and start addressing any issues.

Suspecting this is very similar to what was done for spartacus.

muffinresearch:
This just adds a no-op gettext func in utils and runs the extraction with grunt-i18n-abide.


public/jsx/components/card-form.jsx
```diff
@@ -20,7 +21,7 @@ module.exports = React.createClass({
         {
           'data-braintree-name': 'cardholder_name',
           'id': 'cardholder',
-          'placeholder': 'Card holder name',
+          'placeholder': gettext('Card holder name'),
```
andymckay: I don't see this in @brampitoyo mocks https://github.com/mozilla/payments/pull/55/files

andymckay:
lgtm

public/jsx/components/card-form.jsx
muffinresearch: Yeah, I need to do a pass to make sure things match up 100%

PR #58, opened 2015-05-22
Bring form-serialization into router (fixes #59, fixes #56)

### LINKED ISSUE #56: Prevent user from clicking Subscribe button twice
To prevent double charging, we should see if we can do something that prevents the user from clicking the Subscribe button twice accidentally. There could be some server logic to prevent this as well in https://github.com/mozilla/payments-service/issues/25

### LINKED ISSUE #59: Tokenise card and send to API

muffinresearch:
I ended up doing a lower level card data serialization so that we could have more control over the submission.

* Do the serialization with the braintree client and post to the form
* UX is me just winging it. It'll be fixed up to match the mocks in due course. (thought svg path animating that tick might be nice). 

TODOs in future branches:
* Add lots and lots of tests and error handling. :bomb: 
* Pass real data into the completion page.
* Add a button to exit the flow.

### DIFF
--- a/public/jsx/app.jsx
+++ b/public/jsx/app.jsx
@@ -6,6 +6,7 @@ var Route = Router.Route;
 var RouteHandler = Router.RouteHandler;
 
 var CardDetails = require('views/card-details');
+var CompletePayment = require('views/complete-payment');
 var Login = require('views/login');
 
 var App = React.createClass({
@@ -21,7 +22,8 @@ var App = React.createClass({
 var routes = (
   <Route handler={App}>
     <Route name="login" path="/" handler={Login}/>
-    <Route name="card-form" path="/payment-card/" handler={CardDetails}/>
+    <Route name="card-form" path="/payment/card/" handler={CardDetails}/>
+    <Route name="complete" path="/payment/complete/" handler={CompletePayment}/>
   </Route>
 );
 
--- a/public/jsx/components/card-form.jsx
+++ b/public/jsx/components/card-form.jsx
@@ -1,8 +1,11 @@
 'use strict';
 
-var React = require('react');
-var MaskedInput = require('react-maskedinput');
+var $ = require('jquery');
 var CardValidator = require('card-validator');
+var MaskedInput = require('react-maskedinput');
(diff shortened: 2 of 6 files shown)


public/jsx/app.jsx
```diff
@@ -6,6 +6,7 @@ var Route = Router.Route;
 var RouteHandler = Router.RouteHandler;
 
 var CardDetails = require('views/card-details');
+var Complete = require('views/complete');
```
kumar303: nit: this makes sense as a URL but it's a little vague on its own. Maybe ``CompletePay`` and ``views/complete-pay`` ?

public/jsx/components/card-form.jsx
```diff
@@ -54,6 +62,10 @@ module.exports = React.createClass({
     };
   },
 
+  contextTypes: {
+    router: React.PropTypes.func,
+  },
```
kumar303: This along with the mixins seem like something that would need to be in all of our components. If that's the case, could we make a base component to inherit from?

public/jsx/components/card-form.jsx
```diff
@@ -98,6 +110,39 @@ module.exports = React.createClass({
...
+        $.ajax({
+          data: {
+            pay_method_nonce: nonce,
+            plan_id: 'mozilla-concrete-brick',
+          },
+          url: '/api/braintree/subscriptions/',
+          method: 'post',
+          dataType: 'json',
+        }).then(function(subscribeData) {
+          console.log('successfully subscribed?', subscribeData);
```
kumar303: As far as I know this response won't have any content (unless we think of something) so the log message could just be generic.

kumar303:
Pretty slick! The direct approach with the braintree tokenize function seems like a good idea.

r+wc

public/jsx/app.jsx
muffinresearch: yep, good call.

public/jsx/components/card-form.jsx
muffinresearch: Good point. We should definitely consider doing something like that.

PR #67, opened 2015-05-27
Move to using Payments Bot for travis github pushes

muffinresearch:
* Add new encrypted secrets
* Point at different branch (will delete the old docker one in due course)
* Add test for the gh-pages config silent option.


andymckay:
r+

PR #68, opened 2015-05-28
Remove card holder name (fixes #63)

### LINKED ISSUE #63: Remove card holder name
The mocks don't have that on and we don't pass the data through to braintree.

muffinresearch:
I love how easy this was.


andymckay:
What the, really?

r+

PR #72, opened 2015-05-29
Add tooltips for client-side errors (fixes #49, fixes #69)

### LINKED ISSUE #49: Create React-based error message overlays for cc form

### LINKED ISSUE #69: Show error messages for client-side errors

muffinresearch:


public/jsx/components/input-error.jsx
```diff
@@ -0,0 +1,27 @@
...
+    optModifier: React.PropTypes.oneOf(['center', 'right', 'left']),
+    text: React.PropTypes.string.isRequired,
+  },
+
+  render: function() {
+    var { text, ...toolTipAttrs } = this.props;
+
+    var errorClassList = ['tooltip'];
+    errorClassList.push(this.props.optModifier || 'left');
+    var errorClass = errorClassList.join(' ');
```
andymckay: Could you use classSet here? https://facebook.github.io/react/docs/class-name-manipulation.html

andymckay:
r+wc

public/jsx/components/input-error.jsx
muffinresearch: Could do - I used it earlier when switching to react. For whatever reason it's not built-in now, it's another lib. Given this is small I'll just do it manually. If I get to the point that this is becoming a repeating pattern it'll be worth adding the classSet dep

PR #77, opened 2015-06-03
Handle api errors (fixes #75, fixes mozilla/payments#45)

### LINKED ISSUE #75: Refactor react form instances to better handle errors and custom field logic

muffinresearch:
* Splits up the form into more components with the parent form controlling state
* Each component is therefore more readily testable in its own right + it's clearer that each component deals with the things that affect it's appearance based on props passed down from the parent.
* The form render func is now quite a bit less sprawling and it's a bit easier to see what's responsible for what.
* The API error handing for form level errors is now handled.

Now this unlocks easier testing. I'll do that tomorrow and work out what to do with generic level errors for processing issues.


public/jsx/components/card-form.jsx
```diff
@@ -138,135 +110,122 @@ module.exports = React.createClass({
...
+      generic: [],
+    };
+    if (errors.error_response && errors.error_response.braintree) {
+      var apiErrors = errors.error_response.braintree;
+      // Iterate over the error object and create a new data
+      // structure keyed by field or otherwise push onto
+      // a list of generic errors.
+      Object.keys(apiErrors).forEach(function(key) {
+        var errorList = apiErrors[key];
+        for (var i = 0; i < errorList.length; i += 1) {
```
kumar303: You can probably use a fancy ES7 iterator here, right? Maye: ``for (var item of errorList) {...``

public/jsx/components/card-form.jsx
```diff
@@ -138,135 +110,122 @@ module.exports = React.createClass({
...
-        if (!isValid) {
-          fieldClassNames.push('invalid');
-          allValid = false;
+  processApiErrors: function(errors) {
+    errors = errors || {};
+    var newErrors = {
+      fields: {},
+      generic: [],
+    };
+    if (errors.error_response && errors.error_response.braintree) {
```
kumar303: There is a possibility that you'll get an error but that ``error_response`` will be an empty object. I think you probably need to handle this case by showing a generic message like "unknown error occurred"

public/jsx/components/card-form.jsx
```diff
@@ -138,135 +110,122 @@ module.exports = React.createClass({
...
+      card: card,
+      expiration: expiration,
+      cvv: cvv,
+    };
 
-      if (fieldClassNames.length) {
-        fieldClass = fieldClassNames.join(' ');
+    var that = this;
+    // Iterating over the fieldProps augment the initial
+    // props with dynamic updates.
```
kumar303: This sentence reads awkwardly to me.

public/jsx/components/card-form.jsx
```diff
@@ -138,135 +110,122 @@ module.exports = React.createClass({
...
+      var fieldData = fieldProps[key];
+      var valData = fieldData.validator(val);
+      var cardData = valData.card || {};
+      var id = fieldData.id;
+      fieldData.type = fieldData.type || 'text';
+      fieldData.hasVal = val.length > 0 || false;
+      fieldData.isValid = valData.isValid === true;
+      fieldData.showError = !valData.isValid && !valData.isPotentiallyValid;
+      fieldData.cardType = cardData.type ? cardData.type : null;
+      fieldData.onChangeHandler = that.handleChange;
```
kumar303: why not just define this handler on the CardInput component?

public/jsx/components/card-form.jsx
```diff
@@ -138,135 +110,122 @@ module.exports = React.createClass({
...
+      fieldData.pattern = patternData.pattern;
+      fieldData.placeholder = patternData.placeholder;
+      fieldData.label = patternData.label || patternData.placeholder;
+
+      // Handle field level API errors.
+      if (that.state.errors.fields && that.state.errors.fields[id]) {
+        fieldData.isValid = false;
+        fieldData.showError = true;
+        fieldData.errorMessage = that.state.errors.fields[id][0].message;
+        formIsValid = false;
```
kumar303: it feels like a lot of ``fieldData`` logic in here could be moved to render() of the CardInput component, no?

public/jsx/components/submit-button.jsx
```diff
@@ -0,0 +1,37 @@
...
+    // button to be automagically disabled.
+    if (this.props.showSpinner) {
+      isDisabled = true;
+    }
+
+    return (
+      <button {...buttonAttrs}
+              className={buttonClassNames}
+              disabled={isDisabled}
+              type="submit">Subscribe</button>
```
kumar303: since you're already using gettext, maybe localize 'Subscribe' too?

public/jsx/components/card-form.jsx
```diff
@@ -69,54 +71,25 @@ module.exports = React.createClass({
...
-    },
-  },
-
-
-  handleChange: function(index, e) {
-    var stateChange = {};
-    var val = e.target.value;
-    stateChange[e.target.id] = val;
-    this.setState(stateChange);
+  cardPatterns: cardPatterns,
```
kumar303: I don't see a whole lot of benefit in moving this definition to a separate file. It's not shared logic. It doesn't stand on its own. The downside by putting it in a separate file is that when reading the card-form code you have to open up a new file to read this part.

kumar303:
I like the separation but I feel like more logic could be pushed into CardInput somehow. I don't have concrete suggestions for that though so maybe it has to be the way it is.

r+wc

kumar303:
One thing I don't understand about React is the scope of ``setState()``. Is that state seen by all components? If it's only seen per component then maybe you can't push as much stuff as I thought into CardInput.

public/jsx/components/card-form.jsx
muffinresearch: Then I'd have to do it 3 times.

public/jsx/components/card-form.jsx
muffinresearch: You can't do it that way because state and validation is controlled by the parent. A component is a product of the props passed into it.

muffinresearch:
> One thing I don't understand about React is the scope of setState(). Is that state seen by all components? If it's only seen per component then maybe you can't push as much stuff as I thought into 
CardInput.

The setState calls in the CardForm are setting state only on the form. The form state propagates to the chidren only through props.

The best way I found to think of the data flow is that it flows from the top of the tree flowing down from parents to children.

Each component is the product of it's props. In the form case the form state is managed in the form which "owns" the CardInput fields. This way the parents render method controls all aspects of the children's display by virtue of the props it sends to them. Without this it would be hard to tie aspects of the form together because as an example the CVV field needs to change when the card type changes (although there's a bug in how the masking works that means the pattern doesn't update correctly.

In a similar fashion to know if the button should be "told" to be disabled, it's the form that holds the dynamic state based on the validation of all the relevant fields and based on if the form is being submitted.

Let me know if any of that's not clear - would be happy to discuss it more. I went the route in the beginning of thinking more logic should live per input, but it's not possible to have things work that way - hopefully the reasons above demonstrate this layout.

public/jsx/components/submit-button.jsx
muffinresearch: Yep, good catch.

public/jsx/components/card-form.jsx
muffinresearch: Yeah, fair point. I did this early on in trying to clean things up. I might see if it can be merged with the other form object instead.

public/jsx/components/card-form.jsx
muffinresearch: Ah ok - I'll deal with that when I cover the more generic processing errors.

public/jsx/components/card-form.jsx
muffinresearch: I'll look into that. Still getting used to all the new possibilities. It's a brave new world!

public/jsx/components/card-form.jsx
muffinresearch: Yeah I left that there just to test you, honest!

muffinresearch:
So I looked a bit more and managed to push some more logic into CardInput where it made sense. The pattern pieces are derived solely from the cardType so cardTyps is now set as state on the `<CardForm/>` and passed into the `<CardInput/>`s and they then update the pattern as needed.

I also move the fieldProp changes into handleChange. This way React does less work the data passed as props to the various `<CardInput />`s is only updated when needed rather than updating the data for *all* fields on render.

Note: `fieldProps` is deliberately *not* `props` because it's the data passed to the children, (which becomes the props of the children) rather than being props of the form. React doesn't allow a Component to change its own props.

muffinresearch:
I'm going to land this as is, so I can do stuff that depends on it. Feel free to take a look at the last commit and let me know if you have any comments. We can also take a deeper dive into anything that doesn't make sense in this afternoon's stand-up.

PR #79, opened 2015-06-04
Show card declined error (fixes #45)

### LINKED ISSUE #45: Use build/dist dir for built css/js

muffinresearch:


public/jsx/components/card-form.jsx
```diff
@@ -158,24 +164,20 @@ module.exports = React.createClass({
...
+          fieldData.isValid = false;
+          fieldData.showError = true;
+          fieldData.errorMessage = fieldData.errors[errorData.error];
         }
       });
     }
-    return newErrors;
+    this.setState({
+      isSubmitting: false,
+    });
```
kumar303: setting the state ``isSubmitting: false`` seems like something that should not happen only when there are errors. It should happen at the end of any Ajax response regardless of success/fail. Could you not put it in ``$.ajax(...).always(...)`` ?

muffinresearch: Yep - that would work. Bear in mind it's really a no-op for success due to the route change.

public/jsx/components/card-form.jsx
```diff
@@ -158,24 +164,20 @@ module.exports = React.createClass({
...
-          }
+        console.log('API ErrorMessage: ' + JSON.stringify(apiErrors[key]));
+        var errorData = that.errorKeyToFieldMap[key] || {};
+        var field = errorData.field;
+        if (field) {
+          var fieldData = that.fieldProps[field];
+          fieldData.isValid = false;
+          fieldData.showError = true;
+          fieldData.errorMessage = fieldData.errors[errorData.error];
         }
```
kumar303: This needs an else branch to show any un-mapped field errors somehow. I think a good fallback would be to stick the error into ``__all__`` with a warning in the console showing that we forgot to map the key.

tests/test.card-form.jsx
```diff
@@ -59,4 +59,47 @@ describe('Card Details', function() {
...
+        id: 'expiration',
+      },
+    });
+    var expiration = findByClass(cardForm, 'expiration');
+    assert.include(expiration.props.className, 'invalid');
+    var expirationError = findByClass(expiration, 'tooltip');
+    assert.ok(TestUtils.isCompositeComponent(expirationError));
+  });
+
+  it('shows a cvv message on a api error', function() {
```
kumar303: nit: on an api...

public/jsx/components/card-form.jsx
```diff
@@ -135,22 +135,28 @@ module.exports = React.createClass({
...
+  },
+
   processApiErrors: function(errors) {
-    errors = errors || {};
-    var newErrors = {
-      fields: {},
-      generic: [],
-    };
     var that = this;
     if (errors.error_response && errors.error_response.braintree) {
```
kumar303: is there an issue filed to handle non-braintree errors too? Some mozilla errors could be about seller/product configuration or lookup errors (which we always seem to get when refreshing dev databases, sigh)

public/jsx/components/card-form.jsx
kumar303: mm, yeah. It just seems weird to be on error only. I guess it's ok if the route change makes it moot.

kumar303:
r+wc

public/jsx/components/card-form.jsx
muffinresearch: I was originally planning to fall-back to the larger form level error if this was the case. Because we wouldn't know what that realated to.

public/jsx/components/card-form.jsx
muffinresearch: Those would be the form-level unexpected errors I think as above. I'll work that in when I get to it.

muffinresearch:
I've raised https://github.com/mozilla/payments-ui/issues/80 for un-anticipated errors so I'll cover the remaining unfixed fallbacks there.

PR #86, opened 2015-06-08
Display existing cards (fixes #73)

### LINKED ISSUE #73: List existing cards
To support mozilla/payments#65 we'll need to have a web view that supports existing cards and shows the existing cards and let's a user select that.

For purposes of the first pass let's not worry about editing the card in the buy view, just completing a purchase using an existing card.

muffinresearch:
This branch:

* Displays existing cards and allows the user to select a card from the list
* Sends the user to the card form if the card list is empty (We'll optimize this post-login later)
* Sends the user to the card list when they have cards.
* Clean up of test objects
* Added some helpers - because the React testutils methods areVeryLongWindedAndFillUpFarTooMuchSpace

TODOs to come later:

* More tests
* Checking of the card type handling - this is currently slightly blocked by the inability to add amex due to the input masking not handling pattern updates.

### DIFF
--- a/public/jsx/app.jsx
+++ b/public/jsx/app.jsx
@@ -6,6 +6,7 @@ var Route = Router.Route;
 var RouteHandler = Router.RouteHandler;
 
 var CardDetails = require('views/card-details');
+var CardListing = require('views/card-listing');
 var CompletePayment = require('views/complete-payment');
 var Login = require('views/login');
 
@@ -23,7 +24,10 @@ var routes = (
   <Route handler={App}>
     <Route name="login" path="/" handler={Login}/>
     <Route name="card-form" path="/payment/card/" handler={CardDetails}/>
-    <Route name="complete" path="/payment/complete/" handler={CompletePayment}/>
+    <Route
+      name="complete" path="/payment/complete/" handler={CompletePayment}/>
+    <Route
+      name="card-listing" path="/payment/card-list/" handler={CardListing}/>
   </Route>
 );
 
--- a/public/jsx/components/card-icon.jsx
+++ b/public/jsx/components/card-icon.jsx
@@ -4,7 +4,7 @@ var classNames = require('classnames');
 var React = require('react');
 
 
-var CardIcon = React.createClass({
+module.exports = React.createClass({
   displayName: 'CardIcon',
 
   propTypes: {
@@ -14,6 +14,7 @@ var CardIcon = React.createClass({
       'discover',
       'jcb',
       'maestro',
+      'mastercard',
       'master-card',
       'visa',
     ]),
@@ -37,5 +38,3 @@ var CardIcon = React.createClass({
     return cardType ? <span className={cardClassName} /> : null;
   },
 });
-
-module.exports = CardIcon;
--- a/public/jsx/components/card-item.jsx
+++ b/public/jsx/components/card-item.jsx
@@ -1,0 +1,36 @@
+'use strict';
+
+var React = require('react');
+var CardIcon = require('components/card-icon');
+
+
+module.exports = React.createClass({
+
+  displayName: 'CardItem',
+
+  propTypes: {
+    id: React.PropTypes.number.isRequired,
+    onChangeHandler: React.PropTypes.func.isRequired,
+    resource_uri: React.PropTypes.string.isRequired,
+    truncated_id: React.PropTypes.string.isRequired,
+    type_name: React.PropTypes.string.isRequired,
+  },
+
+  render: function() {
+    var cardType = this.props.type_name.toLowerCase();
+    var cardText = '●●●● ●●●● ●●●● ' + this.props.truncated_id;
+    var inputId = 'card-' + this.props.id;
+
+    return (
+      <li className={'card-item ci-' + this.props.id}>
+        <CardIcon cardType={cardType} />
+        <input id={inputId} type="radio"
+               onChange={this.props.onChangeHandler}
+               value={this.props.resource_uri} name="card" />
+        <label htmlFor={inputId}
+               className="text">{cardText}</label>
+      </li>
+    );
+  },
+
+});
--- a/public/jsx/components/card-list.jsx
+++ b/public/jsx/components/card-list.jsx
@@ -1,0 +1,84 @@
+'use strict';
+
+var $ = require('jquery');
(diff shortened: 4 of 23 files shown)


public/jsx/app.jsx
```diff
@@ -23,7 +24,10 @@ var routes = (
   <Route handler={App}>
     <Route name="login" path="/" handler={Login}/>
     <Route name="card-form" path="/payment/card/" handler={CardDetails}/>
-    <Route name="complete" path="/payment/complete/" handler={CompletePayment}/>
+    <Route
+      name="complete" path="/payment/complete/" handler={CompletePayment}/>
+    <Route
+      name="card-listing" path="/payment/card-list/" handler={CardListing}/>
```
kumar303: I think we should get in the habit of calling these payment methods (I've been going with ``paymethods`` for short). It doesn't have to be in this commit but maybe later. The reason is we'll likely have more than just credit cards at some point.

kumar303: but the URLs should be easy to change later since nothing will rely on the URLs.

public/jsx/components/card-icon.jsx
```diff
@@ -14,6 +14,7 @@ var CardIcon = React.createClass({
       'discover',
       'jcb',
       'maestro',
+      'mastercard',
       'master-card',
```
kumar303: Do we need both ``mastercard`` and ``master-card`` ?

public/jsx/components/card-item.jsx
```diff
@@ -0,0 +1,36 @@
...
+    id: React.PropTypes.number.isRequired,
+    onChangeHandler: React.PropTypes.func.isRequired,
+    resource_uri: React.PropTypes.string.isRequired,
+    truncated_id: React.PropTypes.string.isRequired,
+    type_name: React.PropTypes.string.isRequired,
+  },
+
+  render: function() {
+    var cardType = this.props.type_name.toLowerCase();
+    var cardText = '●●●● ●●●● ●●●● ' + this.props.truncated_id;
```
kumar303: does it matter that different number formats will be different? Amex would not have four numbers in a row, for example. Is there a way to just apply blanks to the pattern formatter?

public/jsx/components/card-list.jsx
```diff
@@ -0,0 +1,84 @@
...
+var Navigation = require('react-router').Navigation;
+
+var CardItem = require('components/card-item');
+var SubmitButton = require('components/submit-button');
+
+var gettext = require('utils').gettext;
+
+
+module.exports = React.createClass({
+  displayName: 'CardList',
```
kumar303: since there is already a ``CardListing`` could this be ``CardSelect`` maybe?

public/jsx/views/card-listing.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+  contextTypes: {
+    router: React.PropTypes.func,
+  },
+
+  render: function() {
+    if (this.state.cards) {
+     return <CardList cards={this.state.cards}/>;
+    } else {
+     return <Spinner text={gettext('Reticulating splines')}/>;
```
kumar303: and which splines would those be?

public/jsx/views/card-listing.jsx
```diff
@@ -0,0 +1,71 @@
...
+
+  propTypes: {
+    apiSource: React.PropTypes.string.isRequired,
+  },
+
+  mixins: [Navigation],
+
+  getDefaultProps: function() {
+    return {
+      apiSource: '/api/braintree/mozilla/paymethod/',
```
kumar303: why does this need to be a property? Is it configurable?

tests/helpers.js
```diff
@@ -30,4 +32,13 @@ module.exports = {
       },
     },
   },
+
+  findByClass: function(component, className){
+    return TestUtils.findRenderedDOMComponentWithClass(component, className);
```
kumar303: ha, wow. That is a mouthful.

kumar303: you could maybe use the ``fn.apply(arguments.slice())`` trick but I forget how it works.

tests/test.card-list.jsx
```diff
@@ -0,0 +1,107 @@
...
+      var cardIcon = helpers.findByClass(this.CardList, 'cctype-' + cardType);
+      assert.notEqual(cardIcon, null);
+    };
+  }
+
+  cards.forEach(function(card) {
+    it('Displays each card type ' + card, testCardType(card));
+  });
+
+  it('State is updated when card is selected', function() {
```
kumar303: these are generally meant to read like sentences, maybe something like "it('updates state when card is selected', ...)"

kumar303:
Looks cool, r+wc

public/jsx/components/card-icon.jsx
muffinresearch: Yeah, beacause the card validation returns master-card but the card type in the card list has MasterCard (which is then lower-cased). If that's from our end we can unify things in the service - (I didn't get around to checking) but since the mapping makes master-card into mastercard this will work fine.

public/jsx/components/card-item.jsx
muffinresearch: Yeah, that's a todo - I can't test other types that change patterns yet due to the pattern bug in the lib we're using as I mentioned in the description above.

public/jsx/components/card-list.jsx
muffinresearch: Yep, that's a plan - I'll go for CardChoice as that's a bit more accurate for a set of radio buttons.

public/jsx/views/card-listing.jsx
muffinresearch: I guess not, no - I'll rip those out. They probably should all get move to a settings files at some point.

tests/test.card-list.jsx
muffinresearch: Yep, my bad.

public/jsx/views/card-listing.jsx
muffinresearch: Aha, cool you found my placeholder text.

tests/helpers.js
muffinresearch: Good idea.

PR #89, opened 2015-06-09
Fix growing button (fixes #78)

### LINKED ISSUE #78: Fix button from growing when spinner is shown
The submit button shouldn't grow when the spinner is shown.

The cause is the spinner icon.

muffinresearch:


muffinresearch:
Landing r=trivial

PR #90, opened 2015-06-09
Fix for pushState error (fixes #83)

### LINKED ISSUE #83: Find workaround for pushState error
Am seeing this when force-reloading the pay.dev site and trying to purchase. 

https://bugzilla.mozilla.org/show_bug.cgi?id=1003100

Worst case we can turn HTML5 history off because we don't really need the server page reloading at all since reloading the hosted client would clear the modal anyway.

muffinresearch:
See also mozilla/payments-client#16

### DIFF
--- a/public/jsx/app.jsx
+++ b/public/jsx/app.jsx
@@ -33,7 +33,7 @@ var routes = (
 
 module.exports = {
   init: function() {
-    Router.run(routes, Router.HistoryLocation, function(Root) {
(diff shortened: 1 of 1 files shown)


kumar303:
It's a shame we have to do this but at least no one should be looking at (and caring about) the URLs. Maybe add a comment in the code linking back to the bug so we remember in the future why we're not using pushState?

r+wc

muffinresearch:
Yep it'll be in the releases details for the client.

Issue #84, opened 2015-06-05
replace jquery with fetch polyfill


kumar303:
I don't think we're using JQuery for anything other than its XHR abstraction, right? We could replace it with the ES6 fetch polyfill: https://github.com/github/fetch :
````
fetch('/users.json')
  .then(function(response) {
    return response.json()
  }).then(function(json) {
    console.log('parsed json', json)
  }).catch(function(ex) {
    console.log('parsing failed', ex)
  })
````

muffinresearch:
Nothing I can't get rid of easily. :+1:

muffinresearch:
Would be good to start graphing the bundle sizes. If only to see it drop considerably when we remove Jquery :)

kumar303:
+1

we would need the obligatory http://arepaymentsslimyet.com of course

kumar303:
This could also be our chance to add a networking layer that can be easily mocked/stubbed for testing. For example, we want to be able to make assertions about how components use their state to do API requests and how they set state based on responses.

PR #91, opened 2015-06-09
Store and send CSRF tokens to the API

kumar303:
Part of https://github.com/mozilla/payments-service/issues/46

I can't think of an easy way to test this. Any ideas?

Going forward, I think we could make an API abstraction (perhaps around fetch, see https://github.com/mozilla/payments-ui/issues/84). If we design the abstraction with testability in mind we could stub it out in tests so that components are fed mock return data. This would allow the tests to make assertions about what components do with the data.

### DIFF
--- a/public/jsx/views/login.jsx
+++ b/public/jsx/views/login.jsx
@@ -29,9 +29,17 @@ module.exports = React.createClass({
       method: 'post',
       url: '/api/auth/sign-in/',
       context: this,
-    }).then(function() {
+    }).then(function(data) {
       if (this.isMounted()) {
         this.setState({'is_logged_in': true}); // eslint-disable-line
+
+        console.log('setting CSRF token for subsequent requests:', data.csrf_token);
+        $.ajaxSetup({
+          headers: {
+            "X-CSRFToken": data.csrf_token,
+          },
+        });
+
         this.transitionTo('card-listing');
       }
     }).fail(function() {


public/jsx/views/login.jsx
```diff
@@ -29,9 +29,17 @@ module.exports = React.createClass({
       method: 'post',
       url: '/api/auth/sign-in/',
       context: this,
-    }).then(function() {
+    }).then(function(data) {
       if (this.isMounted()) {
         this.setState({'is_logged_in': true}); // eslint-disable-line
+
+        console.log('setting CSRF token for subsequent requests:', data.csrf_token);
+        $.ajaxSetup({
```
muffinresearch: Presumably this is going to add the header and not unset the dataType setting? We're already running it here: https://github.com/mozilla/payments-ui/blob/master/public/js/main.js#L7

muffinresearch:
I'm ok with not worrying about the tests for now give that if this isn't sent correctly the API will lock us out. We can definitely do something down the line (along the lines of what you're suggesting).

r+wc

public/jsx/views/login.jsx
kumar303: yeah, it looks like something you can run multiple times and not override old values. I'll test it and see. It's treating jQuery like a static store of saved data. If I were to do it some other way we'd still need the same concept of keeping a static store of data in memory.

kumar303:
depends on https://github.com/mozilla/payments-service/pull/52

Issue #93, opened 2015-06-10
Select the card if there's only one


muffinresearch:

muffinresearch:
As a small pointer - this is where the cards are pushed onto a list: https://github.com/mozilla/payments-ui/blob/master/public/jsx/components/card-choice.jsx#L66

PR #101, opened 2015-06-12
choose if one card

andymckay:
* this might be right, but need to do some tests
* and think about it

### DIFF
--- a/public/jsx/components/card-choice.jsx
+++ b/public/jsx/components/card-choice.jsx
@@ -22,7 +22,8 @@ module.exports = React.createClass({
   getInitialState: function() {
     return {
       isSubmitting: false,
-      card: null,
+      card: this.props.cards.length === 1 ? this.props.cards[0].resource_uri : null,
+
     };
   },
 
@@ -64,7 +65,8 @@ module.exports = React.createClass({
     for (var i = 0; i < cardData.length; i += 1) {
       var card = cardData[i];
       cardList.push(<CardItem {...card} key={'ci-' + i}
-                     onChangeHandler={this.handleCardChange} />);
+                     onChangeHandler={this.handleCardChange}
+                     checked={this.state.card} />);
     }
 
     var formIsValid = this.state.card !== null;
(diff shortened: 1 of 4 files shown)


tests/test.card-choice.jsx
```diff
@@ -105,3 +105,32 @@ describe('Card Choice', function() {
...
+
+    it('Has the card selected when just one card', function() {
+        var input = helpers.findByTag(this.CardChoice, 'input');
+        assert.equal(input.getDOMNode().getAttribute('checked'), '');
+    });
+
+    it('Has the submit button enabled when just one card', function() {
+      var submitButton = helpers.findByTag(this.CardChoice, 'button');
+      assert.equal(submitButton.getDOMNode().getAttribute('disabled'), null);
+    });
```
kumar303: It would be cool to test that pressing submit passes the selected card to the API but we need a mockable API layer first: https://github.com/mozilla/payments-ui/issues/84#issuecomment-111586375

kumar303:
Nice and simple, r+

public/jsx/components/card-choice.jsx
```diff
@@ -64,7 +65,8 @@ module.exports = React.createClass({
     for (var i = 0; i < cardData.length; i += 1) {
       var card = cardData[i];
```
kumar303: This code runs within a render function which means it gets called for any state change. I thought of a situation where a person has multiple cards, selects one, submits it, and gets a gateway error like 'card declined'. In that case, the form will be re-rendered and ``this.state.card`` will be set to the one he selected. So, you need to do the checked feature more like this:

````
var checked = this.state.card && card.resource_uri === this.state.card;
cardList.push(<CardItem ... checked={checked} />);
````

You may want to add another test (in the multiple cards test case) to make sure the right one gets the ``checked`` property.

kumar303:
Actually, heh, I noticed something in the checked logic that needs fixing.

andymckay:
fixes #93

public/jsx/components/card-choice.jsx
andymckay: Good call, wrong card is selected.

andymckay:
I cannot figure out a test for this that works, I tried getting the renderedDOM components and calling Simulate and it never seems to work. Bit lost now, but hey it works.

andymckay:
Well there's a test that the props update, better than nothing.

kumar303:
r+

PR #102, opened 2015-06-12
show the description not the en

andymckay:
before 

screenshot 2015-06-12 11 41 56

after

screenshot 2015-06-12 11 41 13

### DIFF
--- a/public/jsx/components/product-detail.jsx
+++ b/public/jsx/components/product-detail.jsx
@@ -25,7 +25,7 @@ module.exports = React.createClass({
     // l10n. Also let's add a localized currency symbol.
     return (
       <div className="product">
-        <h1 className="title">{productData.id}</h1>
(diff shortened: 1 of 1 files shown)


public/jsx/components/product-detail.jsx
```diff
@@ -25,7 +25,7 @@ module.exports = React.createClass({
     // l10n. Also let's add a localized currency symbol.
     return (
       <div className="product">
-        <h1 className="title">{productData.id}</h1>
+        <h1 className="title">{productData.description.en}</h1>
```
kumar303: It seems to me like we're missing something from the config. Don't we want a shorter product name here? A "title" ? A description sounds like it could be a brief summary but maybe the term "description" is misleading.

kumar303:
r+wc

andymckay:
That's a payments config change, but its an improvement over `id`.

PR #103, opened 2015-06-12
transparent background and pull logo

andymckay:
screenshot 2015-06-12 15 33 20

### DIFF
--- a/public/index.html
+++ b/public/index.html
@@ -9,7 +9,7 @@
   </head>
   <body>
     <main>
-      <div id='view'>
+      <div id="view">
         <div class="spinner-cont">
           <div class="spinner"/>
         </div>
--- a/public/jsx/app.jsx
+++ b/public/jsx/app.jsx
@@ -10,6 +10,8 @@ var CardListing = require('views/card-listing');
 var CompletePayment = require('views/complete-payment');
 var Login = require('views/login');
 
+var products = require('products');
+
 var App = React.createClass({
 
   displayName: 'App',
@@ -28,8 +30,14 @@ var App = React.createClass({
   },
 
   render () {
+    var img = {
+        backgroundImage: 'url(' + products[this.state.productId].img + ')',
+    };
     return (
-      <RouteHandler productId={this.state.productId} />
+      <main>
+        <div id="logo" style={img}></div>
+        <RouteHandler productId={this.state.productId} />
+      </main>
     );
   },
 });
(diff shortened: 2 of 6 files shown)


andymckay:
also requires https://github.com/mozilla/payments-client/pull/17 not perfect but getting there

public/jsx/components/product-detail.jsx
```diff
@@ -17,6 +17,11 @@ module.exports = React.createClass({
     var productId = this.props.productId;
     var productData = products[productId];
 
+    var logo = document.getElementById('logo');
+    if (logo) {
+        logo.style.backgroundImage = 'url(' + productData.img + ')';
+    }
```
kumar303: Altering the DOM directly is a no-no in React because you lose its reactive properties.

What you can do instead (I think) is render the app view with your div element as a wrapper around the ``<RouteHandler>``. In [app.jsx](https://github.com/andymckay/payments-ui/blob/transparent-background/public/jsx/app.jsx#L31) I think you can make render look more like this:
````
render () {
  var productData = products[this.state.productId];
  return (
    <div id="logo" style="background-image: url({productData.img})"></div>
    <main>
      <div id="view">
        <RouteHandler productId={this.state.productId} />
      </div>
    </main>
  );
},
````

**Comment by kumar303** [comment]
r- for the DOM altering. I can try to help with this if it gets tricky.

File: public/jsx/components/product-detail.jsx (hunk 1 above)
Comment (andymckay) [inline public/jsx/components/product-detail.jsx]: Ok I think I see now, thanks.

BTW you can't return more than one element in a render, so the `div` has to go inside the `main` (or similar parent element).

File: public/index.html [hunk 2]
```diff
@@ -8,8 +8,9 @@
     <title>Payments UI</title>
   </head>
   <body>
+    <div id="logo"></div>
```
kumar303: do you still need this now?

public/jsx/app.jsx
```diff
@@ -50,7 +58,7 @@ module.exports = {
   init: function() {
     Router.run(routes, Router.HashLocation, function(Root) {
       // Doesn't actually render anything.
-      React.render(<Root/>, document.body);
+      React.render(<Root/>, document.getElementById('view'));
```
kumar303: don't you want to continue rendering to the ``body`` now that all the HTML is there? Otherwise it seems like you might have multiple ``<main>`` tags

kumar303:
r+wc

public/index.html
andymckay: no, will remove

public/jsx/app.jsx
andymckay: right hmm

PR #104, opened 2015-06-12
Display buyer's email in receipt (fixes #98)

### LINKED ISSUE #98: Display email on subscription receipt page
The subscription receipt page landed in https://github.com/mozilla/payments-ui/pull/94/ but with a placeholder for email. Let's display a real email address. We probably need to adjust the API to expose it.

kumar303:
Some notes:

* I've been trying to wrap my head around [flux](http://facebook.github.io/flux/) but my head is not yet wrapped
* It feels a little bit overkill to use a dispatcher
* the tests probably need to unset the user or to use a mock maybe. I'm not sure. What happens when another test tries to access the UserStore?
* there isn't really a login test yet because of https://github.com/mozilla/payments-ui/issues/84#issuecomment-111586375


andymckay:
this didn't work for (ignore the transparency and other crap from other pull requests):

https://www.dropbox.com/s/2vuqv0uea2f1n5t/Screenshot%202015-06-12%2015.54.45.png?dl=0

andymckay:
Hmm "UserActions: setting current user" Object { email: undefined }

andymckay:
Would help if my payments-service was up to date and returning email addresses! Works now. Yay!

andymckay:
Agreed it feels a bit much, I don't really know though how much we'll need to do this for other things, so adding it in might be the right way to go. It builds, the test pass and it works, I don't feel qualified to say yay or nay beyond that.

I dig the two lines before functions, very python-y.

r+ but with a fair helping of I've got no idea.

kumar303:
Flux still feels overkill right now but this is sort of selling me on it:

> As an application grows, the dispatcher becomes more vital, as it can be used to manage dependencies between the stores by invoking the registered callbacks in a specific order. Stores can declaratively wait for other stores to finish updating, and then update themselves accordingly.

~ http://facebook.github.io/flux/docs/overview.html#a-single-dispatcher

PR #109, opened 2015-06-16
flux-ify the sign-in logic

kumar303:
Before this none of our top-level (routable) components were tested and none of our Ajax code was covered. This applies the flux concept of Action -> Dispatcher -> Store -> Views for flowing data into the app. As a nice side effect it makes things very testable since each part has a reduced set of responsibilities. The testing approach relies on object API specification and mainly proves that each part adheres to the contract but this is still useful.


tests/test.login.jsx
```diff
@@ -0,0 +1,51 @@
...
+    var store = this.Login.UserStore;
+    assert.equal(store.addSetUserListener.firstCall.args[0],
+                 this.Login.onSetUser);
+  });
+
+  it('should clean up user listeners', function() {
+    this.Login.componentWillUnmount();
+    var store = this.Login.UserStore;
+    assert.equal(store.removeSetUserListener.firstCall.args[0],
+                 this.Login.onSetUser);
```
andymckay: So is this asserting that the function onSetUser is being passed?

public/js/user-actions.js
```diff
@@ -13,13 +16,56 @@ function UserActions(localDispatcher) {
...
-    user: user,
-  });
-};
+UserActions.prototype = assign({}, {
+
+  signIn: function(accessToken, opt) {
+    opt = opt || {};
+    if (!opt.jquery) {
+      opt.jquery = $;
+    }
```
chuckharmston: I'm sure there's logic behind it, but passing jQuery around like this feels like an anti-pattern. Shouldn't it be `require`able?

public/js/user-actions.js
```diff
@@ -13,13 +16,56 @@ function UserActions(localDispatcher) {
...
 
 
-UserActions.prototype.setCurrentUser = function(user) {
-  console.log('UserActions: setting current user', user);
-  this.dispatcher.dispatch({
-    actionType: 'set-user',
-    user: user,
-  });
-};
+UserActions.prototype = assign({}, {
```
chuckharmston: I don't understand the point of using assign here. Isn't this literally the exact same as `UserActions.prototype = {`?

public/js/user-actions.js
```diff
@@ -13,13 +16,56 @@ function UserActions(localDispatcher) {
...
+        email: data.buyer_email,
+        is_logged_in: true,
+      });
+
+      console.log('setting CSRF token for subsequent requests:',
+                  data.csrf_token);
+      opt.jquery.ajaxSetup({
+        headers: {
+          'X-CSRFToken': data.csrf_token,
+        },
```
chuckharmston: I don't think this is how you want to go about it:

1. Any future calls to ajaxSetup will smother this.
2. Any attempts to set headers on individual requests will smother this.
3. Its use isn't recommended by jQuery.

Alternately, maintain an object in localstorage or memory that you use as per-request defaults? e.g.

    localStorage.setItem('csrftoken', data.csrf_token);

    // Then in each subsequent call to $.ajax
    $.ajax(assign({}, localStorage.getItem('csrftoken'), {
      method: 'post',
      url: '/api/auth/sign-in/'
    }));

chuckharmston: Actually, you probably just want to jam it in `UserStore`.

public/js/user-actions.js
```diff
@@ -13,13 +16,56 @@ function UserActions(localDispatcher) {
...
+      });
+
+    }).fail(function() {
+
+      console.log('login failed');
```
chuckharmston: Nitpicky style thing, but the blank line on L51 feels awkward.

public/jsx/views/login.jsx
```diff
@@ -12,50 +13,41 @@ var gettext = require('utils').gettext;
...
+  componentWillUnmount: function() {
+    this.UserStore.removeSetUserListener(this.onSetUser);
+  },
+
+  onSetUser: function() {
+    if (!this.isMounted()) {
+      console.log('ignoring events while unmounted');
+      return;
+    }
+    var user = this.UserStore.getCurrentUser();
```
chuckharmston: This will raise an error if the user hasn't been set; not sure the conditional logic in the following lines matters at all. I think you want to try/catch instead?

chuckharmston:
I know payments is taking a different approach to Flux than we are on Marketplace, but this feels needlessly complex: passing the dispatcher singleton around, managing event handlers, etc.

It also strikes me that it would be nearly impossible to make this isomorphic, if that's in your plans.

tests/test.login.jsx
kumar303: yea. It is checking that when unmounted, the event listener is removed

public/js/user-actions.js
kumar303: It is required but we aren't mocking out require statements currently. I agree, it would be nicer to reload the module with mock requires but we haven't set that up yet.

public/js/user-actions.js
kumar303: Correct, it isn't doing anything here. A previous version was inheriting from another object which is why I was using it.

public/js/user-actions.js
kumar303: Yeah, agreed on the header thing. We're going to ditch jquery soon for [fetch](https://github.com/mozilla/payments-ui/issues/84) but I didn't want to change too much in this patch. I like the idea of putting the token on ``UserStore``

public/jsx/views/login.jsx
kumar303: there is a case where the user was set but it will have ``is_logged_in: false``. No exception would be thrown in that case. This happens when the sign-in request fails. If the sign-in request never resolves, this event handler would never be called so it shouldn't encounter the exception case.

kumar303:
It is complex, agreed. I wasn't too happy with the libraries I saw and wanted to maintain the concept of the central dispatcher. I think it makes things easier to test with an explicit decoupling between action, dispatcher, and store but perhaps we could still wrap it in abstractions and keep testability.

I'm definitely thinking of this as a first pass for us to iterate on.

We have no future plans to run this code server side.

Thanks for the review!

public/js/user-actions.js
kumar303: ooh! Dropping in [rewire-webpack](https://www.npmjs.com/package/rewire-webpack) was super easy. That allowed me to fix this, all the ``Thing.Class`` nonsense, and a pile of other crap.

PR #114, opened 2015-06-19
Use saved cards returned at login time (fixes #85)

### LINKED ISSUE #85: Work with buyer's saved payment methods returned in sign-in response
Some code can be simplified in dealing with saved payment methods once this lands: https://github.com/mozilla/payments-service/pull/49

kumar303:


public/js/user-store.js
```diff
@@ -18,6 +18,10 @@ function UserStore(localDispatcher) {
   localDispatcher.register((function(payload) {
     if (payload.actionType === 'set-user') {
       console.log('UserStore: storing current user:', payload.user);
+      //
+      // Example:
+      // {email: 'f@f.com', is_logged_in: true, payment_methods: []}
```
chuckharmston: I like this. Is nice.

article-2450794-05e6dc090000044d-375_634x403

tests/helpers.jsx
```diff
@@ -63,11 +72,27 @@ module.exports = {
     function RouterStub() {}
 
     assign(RouterStub, {
-      transitionTo: function() {},
+        makePath () {},
```
chuckharmston: Hmm...what's the point in stubbing these out? Do they all have default behavior you want to override?

chuckharmston: Ignore me. I somehow missed that we were in tests.

chuckharmston: This seems useful. You might want to make this a requirable helper.

public/jsx/views/card-listing.jsx
```diff
@@ -23,25 +23,16 @@ module.exports = React.createClass({
...
-        } else {
-          console.log('No card data found, showing card form');
-          this.transitionTo('card-form');
-        }
-      }
-    }).fail(function() {
-      // TODO: some error state.
-      console.log('Card list retrieval failed');
-    });
+    var user = UserStore.getLoggedInUser();
```
chuckharmston: Do you want to handle the possible error cases here?

kumar303: I'm not sure yet. Right now it will blow up on the console which is ok. Once we wire up full screen errors (https://github.com/mozilla/payments/issues/70) we'll need to think about some kind of global exception handler. It would be annoying to have to catch every error but maybe we'll have to do that because javascript.

public/js/user-store.js
kumar303: I've been wondering if I should enforce this ``payload.user`` object so it blows up if it's missing some properties. Sigh, types.

chuckharmston:
Looks good!

I'm still trying to wrap my head around the testing landscape, so I'll ping @ngokevin; he's been spending a lot of time thinking about it.

public/jsx/views/card-listing.jsx
ngokevin: might think about using a lighteight flux framework to be able to declaratively connect to stores and not have to depend on singletons. we're currently using flummox, but that has been recently deprecated for redux.

public/js/user-actions.js
```diff
@@ -27,6 +27,7 @@ module.exports = assign({}, {
       console.log('login succeeded');
       this.setCurrentUser({
         email: data.buyer_email,
+        payment_methods: data.payment_methods,
```
ngokevin: jquery isn't very necessary in this day and age. i'd recommend using a specific request library if you need it for requests. we're using superagent (with a promise wrapper)

tests/helpers.jsx
```diff
@@ -172,6 +194,7 @@ module.exports = {
       },
       UserStore: {
```
ngokevin: if you design your components in such a way that they simply receive data through props rather than fetching directly from the store, you wouldn't have to stub out stores. you would simply test your isolated components by passing in props.

public/js/user-actions.js
kumar303: yeah, we're thinking of replacing it with fetch, just haven't got around to it yet. https://github.com/mozilla/payments-ui/issues/84

tests/helpers.jsx
kumar303: The data has to come from somewhere so we're using a flux pattern here to pass stuff around that doesn't fit nicely into a parent->child cascading property thing

public/jsx/views/card-listing.jsx
kumar303: The singletons are easy to mock which is nice. I looked at some flux libs but I decided I wanted to try a roll your own approach first just to see how annoying it was. I think we can simplify it further for sure. This article kind of convinced me that there aren't good libs yet https://medium.com/@dan_abramov/the-evolution-of-flux-frameworks-6c16ad26bb31 Redux was one I looked at it but it doesn't seem production ready yet.

kumar303: oh, right, that article basically points to redux as the answer

ngokevin: yeah that's from the creator of redux :)

ngokevin: i'd recommend one of them, the actions and stores are extremely easy to test in isolation. and they are designed so that a flux instance is passed down from the top of the heirarchy all the way to the bottom. so when you're testing an individual component, you just have to make a fake flux instance. or when you have a component connect to a store, you just have to pass in primitive props.

tests/helpers.jsx
ngokevin: so the common way is to pass a flux instance from top to bottom. and have a utility for the flux instance to connect to a store, which extracts out data to use as props for an inner component.

something like https://github.com/acdlite/flummox/blob/master/docs/docs/api/fluxcomponent.md

kumar303: I see. Interesting.

kumar303:
Thanks for the reviews. I'll probably take a look at redux later to see what porting to it might look like.

PR #117, opened 2015-06-19
Set up rough framework for app errors (fixes #116)

### LINKED ISSUE #116: Set up a framework for showing app errors
As part of https://github.com/mozilla/payments/issues/70, set up an initial framework for catching and displaying app errors. This will not be a complete solution.

kumar303:
Notes:
* The internal error page is not styled yet
* Only the sign in failure is hooked up so far

### DIFF
--- a/public/jsx/app.jsx
+++ b/public/jsx/app.jsx
@@ -5,9 +5,11 @@ var Router = require('react-router');
 var Route = Router.Route;
 var RouteHandler = Router.RouteHandler;
 
+var AppStore = require('app-store');
 var CardDetails = require('views/card-details');
 var CardListing = require('views/card-listing');
 var CompletePayment = require('views/complete-payment');
+var ErrorMessage = require('components/error');
 var Login = require('views/login');
 
 var products = require('products');
@@ -25,8 +27,21 @@ var App = React.createClass({
     var productId = router.getCurrentQuery().product;
 
     return {
+      error: AppStore.getError(),
       productId: productId,
     };
+  },
+
+  componentDidMount: function() {
+    AppStore.addErrorListener(this.onError);
+  },
+
+  componentWillUnmount: function() {
+    AppStore.removeErrorListener(this.onError);
+  },
+
+  onError: function() {
+    this.setState({error: AppStore.getError()});
   },
 
   render () {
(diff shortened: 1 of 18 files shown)


public/jsx/app.jsx
```diff
@@ -25,10 +27,23 @@ var App = React.createClass({
     var productId = router.getCurrentQuery().product;
 
     return {
+      error: AppStore.getError(),
       productId: productId,
     };
   },
 
+  componentDidMount: function() {
+    AppStore.addErrorListener(this.onError);
```
ngokevin: we have a hard dependency between the component and store here. once we port to redux, we should be able to connect to the store, and have the App component simply take ```errors``` as a prop. It'll make it easier to test since you won't have to instantiate a store, simply just ```<App errors={['You must login']}/>```

public/jsx/views/login.jsx
```diff
@@ -21,27 +21,21 @@ module.exports = React.createClass({
...
-    var user = UserStore.getCurrentUser();
-    if (user.is_logged_in) {
-      console.log('current user is logged in; continuing to card-listing');
-      this.transitionTo('card-listing');
-    } else if (!user.is_logged_in) {
-      // TODO: some error state.
-      console.error('user did not log in successfully');
-    }
+    console.log('current user is signed in; continuing to card-listing');
+    this.transitionTo('card-listing');
```
ngokevin: i'd heavily recommend simply rendering the card listing straight here, rather than relying on the router. it will make testing a lot easier.

public/jsx/app.jsx
kumar303: You still need to wrap it (or a parent) with a connector though - same dependency. We'll still need a test that sets up the connector.

public/jsx/views/login.jsx
kumar303: Yeah, I'd like to rethink the way we do redirects with ``transitionTo()``. I guess one nice thing is that we can absolve this view of all responsibilities for showing a card listing. In a redirect model there is only one card-listing view in existence which makes the testing surface smaller.

kumar303:
Thanks for taking a look. I'm going to try porting to redux to see how it feels.

public/jsx/app.jsx
kumar303: one nice thing about redux (as chuck pointed out earlier) is that we won't have to explicitly set up (and test) all the event listeners.

PR #121, opened 2015-06-24
allow users to add in another card

andymckay:
* the longest patch for a single `a` tag evar

### DIFF
--- a/public/scss/inc/vars.scss
+++ b/public/scss/inc/vars.scss
@@ -6,8 +6,10 @@ $header-font: "Fira Sans", Helvetica, Arial, sans-serif;
 $baby-blue: #E0F1FA;
 $light-blue: #0096dc;
 $uniform-blue: #0080D8;
+// Not sure what those other blues are, this is the one that counts.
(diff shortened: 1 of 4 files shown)


public/scss/inc/vars.scss
```diff
@@ -6,8 +6,10 @@ $header-font: "Fira Sans", Helvetica, Arial, sans-serif;
 $baby-blue: #E0F1FA;
 $light-blue: #0096dc;
 $uniform-blue: #0080D8;
+// Not sure what those other blues are, this is the one that counts.
+$the-real-blue: #0095dd;
```
kumar303: why don't you just replace ``$light-blue`` with this?

kumar303:
I'm not totally sure this needs such a thorough test. You could maybe do something instead where you just find the ``Link`` child and make sure it has the right properties. In other words, we can trust that the functionality of ``Link`` itself already works.

r+wc

public/scss/inc/vars.scss
andymckay: different blue!

PR #123, opened 2015-06-25
Port flux concepts to redux

kumar303:
This moves our already existing flux concepts over to the [redux](https://github.com/gaearon/redux/tree) framework. Here are some benefits:

* actions and stores are a lot simpler (they are pure functions without side effects)
* everything is thus easier to test
* no more need to roll our own event handlers to watch for data store changes
* redux offers a way to add dispatcher middleware which could be cool for freezing app state in case of an error


public/jsx/views/card-listing.jsx
```diff
@@ -17,45 +17,45 @@ module.exports = React.createClass({
 
   mixins: [Navigation],
```
muffinresearch: We should kill this I don't think it's needed.

public/jsx/views/card-listing.jsx
```diff
@@ -17,45 +17,45 @@ module.exports = React.createClass({
...
-      );
-    } else {
-      return <Spinner text={gettext('Loading')}/>;
-    }
+    var component = this;
+    return (
+      <Connector select={this.selectData}>
+        {function(result) {
+          if (!result.user.email) {
+            return <Spinner text={gettext('Loading')}/>;
```
muffinresearch: How does this work, is this not an error case?

public/jsx/views/complete-payment.jsx
```diff
@@ -40,17 +34,32 @@ module.exports = React.createClass({
...
 
+  selectData: function(state) {
+    console.log('complete-payment: selectData() firing with', state);
+    return {
+      user: state.user,
+    }
+  },
+
   render: function() {
+    var component = this;
```
muffinresearch: Good idea. Much more clear like this (no pun intended)

public/jsx/views/login.jsx
```diff
@@ -39,7 +43,16 @@ module.exports = React.createClass({
   },
 
   render: function() {
-    return <Spinner text={gettext('Logging in')}/>;
+    var component = this;
+    return (
+      <Connector select={this.selectData}>
```
muffinresearch: this -> component ?

tests/test.app-actions.js
```diff
@@ -1,31 +1,16 @@
...
-
-  });
+describe('appActions', function() {
 
   it('should dispatch error message', function() {
     var error = {debugMessage: 'something happened'};
-    appActions.setError(error.debugMessage);
-    assert.deepEqual(dispatcher.dispatch.args[0][0], {
-      actionType: 'set-app-error',
+    var action = appActions.error(error.debugMessage);
```
muffinresearch: Probably don't need this var if you use more lines. </nitpick>

tests/test.user-actions.js
```diff
@@ -2,35 +2,24 @@
...
+
 var helpers = require('./helpers');
 
 
-describe('UserActions', function() {
+describe('userActions', function() {
 
-  var dispatcher;
-  var appActions;
+  var dispatch;
```
muffinresearch: Maybe call this `dispatchSpy` just to make it clearer lower down?

muffinresearch:
Looks good to me. I don't feel like I get the redux stuff totally yet - but I'm sure it'll sink in once I start using it.

r+wc (pre jetlag review)

public/jsx/views/card-listing.jsx
kumar303: isn't it needed for ``this.transitionTo`` ?

public/jsx/views/login.jsx
```diff
@@ -21,17 +22,20 @@ module.exports = React.createClass({
 
   componentDidMount: function() {
     var { router } = this.context;
-    UserStore.addSignInListener(this.onUserSignIn);
-    UserActions.signIn(router.getCurrentQuery().access_token);
+    var defer = userActions.signIn(router.getCurrentQuery().access_token);
+    defer(reduxConfig.default.dispatch);
```
kumar303: I found out a better way to do this. In the parent I can do:
````
<Login
  accessToken={state.accessToken}
  {...bindActionCreators(userActions, result.dispatch) }
/>
````
which then allows me to do:
````  
componentDidMount: function() {
  this.props.signIn(this.props.accessToken);
},
````

public/jsx/views/card-listing.jsx
kumar303: It's sort of weird. Because this component is listening (via the Connector) to user state change, it's possible that this might get rendered. It's unlikely but possible. Most likely the login view will be rendered and a login occur and the card listing view will get reloaded.

I think this is really where trying to use the router starts to break down. I have another patch in the works that gets rid of the router and takes a more data/state driven approach to rendering views.

public/jsx/views/login.jsx
kumar303: I have this in another patch

PR #129, opened 2015-06-28
Purchase flow without a router

kumar303:
- Removes the [router](https://github.com/rackt/react-router) because I don't think we need it and it's confusing me. Without the router, we can use the Flux Action->Dispatcher->Store->View pattern in a more sane way since all rendering will be in response to data changes.
- Minimizes the number of redux Connectors which makes things easier to test and easier to think about. In other words, there are more "dumb" components that simply receive properties now.
- Makes better use of action binding so that child components can receive simple callable properties (same as above).
- Starts moving some of the purchase Ajax to actions although this is incomplete. Before, the purchase Ajax was hard to test. When we move it all to actions it will be easier to test.


public/jsx/components/card-choice.jsx
```diff
@@ -2,12 +2,13 @@
 
 var $ = require('jquery');
 var React = require('react');
-var Navigation = require('react-router').Navigation;
 
 var CardItem = require('components/card-item');
 var SubmitButton = require('components/submit-button');
 
 var gettext = require('utils').gettext;
+var purchaseActions = require('purchase-actions');
```
muffinresearch: Would it make sense to put all actions in their own dir? So this could be `require('actions/purchase')` for example.

public/jsx/views/card-listing.jsx
```diff
@@ -1,61 +1,37 @@
...
-        }}
-      </Connector>
+      <div className="card-listing">
+        <ProductDetail productId={this.props.productId} />
+        <CardChoice
+          cards={this.props.paymentMethods}
+          productId={this.props.productId}
+        />
+        <a className="card-add bottom-link"
+           onClick={this.props.payWithNewCard}>
```
muffinresearch: I think you need an href attr for this to be keyboard focusable.

public/jsx/app.jsx
```diff
@@ -1,42 +1,55 @@
...
 var ErrorMessage = require('components/error');
 var Login = require('views/login');
+var Purchase = require('views/purchase');
+var userActions = require('user-actions');
 
 var products = require('products');
 
 
+function parseQuery(url) {
+  // TODO: replace with querystring library or something.
```
muffinresearch: I quite like this, which is simple and doesn't get into worrying about nesting: https://github.com/sindresorhus/query-string

tests/test.purchase.jsx
```diff
@@ -0,0 +1,113 @@
...
+      type: actionTypes.USER_SIGNED_IN,
+      user: assign({}, defaultUser, {
+        payment_methods: paymentMethods,
+      }),
+    });
+
+    redux.dispatch(purchaseActions.payWithNewCard());
+
+    // Instead make sure a new card entry form was rendered.
+    var child = TestUtils.findRenderedComponentWithType(
```
muffinresearch: Unnecessary var?

muffinresearch:
Lgtm r+wc. Let's get this in and perhaps we can sync-up when you're in later for a bit of a run-through?

public/jsx/views/card-listing.jsx
kumar303: aha, that's probably why it doesn't show a pointer when you hover over it

public/jsx/components/card-choice.jsx
kumar303: good idea

public/jsx/app.jsx
kumar303: actually, yeah, it's a small amount of code so I guess we can keep it. I'll add a direct test for it.

tests/test.purchase.jsx
kumar303: oh right, I guess it will blow up if not found so the assert below isn't needed.

PR #131, opened 2015-06-29
Configure minified CSS output (fixes #10)

### LINKED ISSUE #10: Add CSS minification

muffinresearch:


config/sass.js
```diff
@@ -2,9 +2,20 @@ module.exports = {
...
     files: {
       'public/dist/main.css': 'public/scss/main.scss',
     },
   },
+  min: {
+    options: {
+      outputStyle: 'compressed'
+    },
+    files: {
+      'public/dist/main.min.css': 'public/scss/main.scss',
```
kumar303: Do we need something to toggle prod and link to this? https://github.com/mozilla/payments-ui/blob/master/public/index.html#L7 Or is that for a later commit?

kumar303:
r+wc

config/sass.js
muffinresearch: Yep later, although this one is less of a performance hit compared to uglify. (which is why we don't want to have to wait for the uglification on watching for example).

PR #132, opened 2015-06-29
Add minified bundle (fixes #11)

### LINKED ISSUE #11: Add JS minification

muffinresearch:


config/webpack.js
```diff
@@ -12,6 +16,17 @@ module.exports = {
     },
     failOnError: true, // don't report error to grunt if webpack find errors
     watch: true, // use webpacks watcher
-    //keepalive: true, // don't finish the grunt task
   },
+  prod: {
+    devtool: 'source-map',
```
kumar303: is there a way to move this up one level so it's inherited by both?

config/webpack.js
```diff
@@ -12,6 +16,17 @@ module.exports = {
...
+    failOnError: true, // don't report error to grunt if webpack find errors
+    output: {
+      path: path.join(__dirname, '../public/dist/'),
+      filename: 'bundle.min.js',
```
kumar303: Just like for css, we probably need a way to link to this in prod, right? https://github.com/mozilla/payments-ui/blob/master/public/index.html#L18

config/webpack.js
```diff
@@ -12,6 +16,17 @@ module.exports = {
...
+    },
+    plugins: [
+      new webpack.optimize.UglifyJsPlugin({minimize: true})
+    ],
+    watch: true, // use webpacks watcher
```
kumar303: same here - maybe there is a way to inherit this as a default for both configs?

kumar303:
r+wc

config/webpack.js
muffinresearch: Yeah, that'll be something needing to be done when we generate the html files (e.g. for l10n too)

config/webpack.js
muffinresearch: This one probably not because tests inherit this.

config/webpack.js
muffinresearch: Yep I reckon.

PR #133, opened 2015-06-29
Move to explicit package versions

muffinresearch:
We've had a bunch of issues with upstreams breaking things noticeable only through tests. This should help prevent us getting packages upgraded implicitly.


kumar303:
Good idea. I think my router removal conflicted.

r+

PR #134, opened 2015-06-29
Update messages.po - support brazilian messages

luhad:
Add support messages in brazilian language

[Review on Reviewable](https://reviewable.io/reviews/mozilla/payments-ui/134)


muffinresearch:
Hi, thanks for the patches.

Could you make the image optimizations a separate PR please? 

iirc the svg should have come through svgo via grunt-svg-sprite. So any manual optimizations will be blown away if we run the command to build it see https://github.com/mozilla/payments-ui/blob/master/config/svg_sprite.js - I suspect tweaking the config might be able to achieve the same changes.

Similarly any other img size reductions should ideally be automated so that we can reproduce them and that updated assets don't lose the optimizations. E.g. using https://github.com/gruntjs/grunt-contrib-imagemin.

muffinresearch:
I've pulled in the locale update. The other commit would need re-proposing as per my previous comment.

PR #138, opened 2015-07-01
Fix up field attrs (fixes #137)

### LINKED ISSUE #137: Ensure name attrs aren't specified on the form
This is so the form values aren't submitted to us if the js submit breaks.
Needs test coverage too.

muffinresearch:
* Remove the braintree data attrs as we're building things up at a lower level.
* Add the autocomplete="off" attr
* Add test for no name attr being present.
* Test for type="tel" and autocomplete

### DIFF
--- a/public/jsx/components/card-form.jsx
+++ b/public/jsx/components/card-form.jsx
@@ -14,6 +14,15 @@ var purchaseActions = require('actions/purchase');
 var reduxConfig = require('redux-config');
 
 
+var defaultFieldAttrs = {
+  'autoComplete': 'off',
+  // inputMode is not yet supported in React.
+  // See https://github.com/facebook/react/pull/3798
+  'inputMode': 'numeric',
+  'type': 'tel',
+};
+
+
 module.exports = React.createClass({
 
   displayName: 'CardForm',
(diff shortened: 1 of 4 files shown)


public/jsx/components/card-form.jsx
```diff
@@ -61,18 +69,17 @@ module.exports = React.createClass({
...
       'pattern': '11/11',
       'placeholder': 'MM/YY',
-      'type': 'tel',
       'validator': CardValidator.expirationDate,
     },
     cvv: {
+      'attrs': defaultFieldAttrs,
+      'autocomplete': 'off',
       'classNames': ['cvv'],
-      'data-braintree-name': 'cvv',
```
andymckay: braintree doesn't need this anymore, or is it added somewhere else?

andymckay: oh i guess that's for drop-in ui?

andymckay:
r+wc

public/jsx/components/card-form.jsx
muffinresearch: Yep - we do the tokenisation at a lower level.

PR #151, opened 2015-07-02
Add csslint (fixes #8)

### LINKED ISSUE #8: Add CSS lint

muffinresearch:
Quite a lot of this is noise so I've trimmed out a lot of that. Because rules are a bit all or nothing (e.g. when you opt-out of a rule you kill it for everything unlike eslint/jshint etc) I've added two modes. The lax mode (run with `grunt csslint:lax`) is what's run as part of tests. This has some errors enforced so that travis will fail if we have duped props. We should probably fix anything that starts popping up here or update the configuration.

The strict mode (run with `grunt csslint:strict`) is to see all the things csslint doesn't like which will be worth running every once in a while to see if we can clean some it up if relevant. I've excluded things that we 100% don't care about from this in the grunt rules.

Whilst ignoring a lot of the rules makes this feel a bit redundant, running csslint did catch a couple of silly things so on that basis it *does* add value.

### DIFF
--- a/public/scss/_base.scss
+++ b/public/scss/_base.scss
@@ -24,7 +24,7 @@ a {
 
   .accepted {
     @include header-font();
-    font-size: 1.75em;
+    font-size: $large-font;
     font-weight: 200;
     padding: 20px 0;
   }
@@ -93,8 +93,8 @@ main {
     & label {
       border: 1px solid transparent;
       display: block;
-      font-size: 18px;
-      font-weight: bold;
+      font-size: $medium-font;
+      font-weight: 700;
       padding: 1em 15px 1em 60px;
       position: relative;
       text-align: right;
--- a/public/scss/_forms.scss
+++ b/public/scss/_forms.scss
@@ -52,7 +52,7 @@ input {
 
   @include respond-to('small') {
     border-radius: $small-border-radius;
-    font-size: $base-font + $media-adjustment;
+    font-size: $base-font;
     height: 40px;
   }
 
--- a/public/scss/_product.scss
+++ b/public/scss/_product.scss
@@ -5,9 +5,9 @@
 
   .title {
     font-size: $medium-font;
-    font-weight: normal;
+    font-weight: 400;
     padding-bottom: 0.35em;
-    margin: 0em;
+    margin: 0;
   }
 
   .price {
--- a/public/scss/_tooltip.scss
+++ b/public/scss/_tooltip.scss
@@ -3,7 +3,7 @@
   transition: 0.1s opacity $short-transition;
   display: block;
   color: #fff;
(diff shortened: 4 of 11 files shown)


public/scss/inc/mixins.scss
```diff
@@ -57,6 +56,6 @@
   (-webkit-min-device-pixel-ratio: 1.3),
   (min-device-pixel-ratio: 1.3),
   (min-resolution: 1.3dppx) {
-    background-image: image-url("#{$filename}@2x.#{$extension}");
```
muffinresearch: An example of something it caught. I'd missed tweaking this line (the mixin image-url was removed in sass3.1)

public/scss/_tooltip.scss
```diff
@@ -6,7 +6,6 @@
   background: #F91200;
   padding: 0.5em 0.75em 0.7em;
   z-index: 10;
-  top: -0.5em;
```
muffinresearch: Dupe selector caught!

public/scss/_product.scss
```diff
@@ -5,9 +5,9 @@
 
   .title {
     font-size: $medium-font;
-    font-weight: normal;
+    font-weight: 400;
     padding-bottom: 0.35em;
-    margin: 0em;
```
muffinresearch: Zeros don't need units.

public/scss/_forms.scss
```diff
@@ -52,7 +52,7 @@ input {
 
   @include respond-to('small') {
     border-radius: $small-border-radius;
-    font-size: $base-font + $media-adjustment;
+    font-size: $base-font;
```
muffinresearch: Killing the media adjustment - we should create another size if we need this.

public/scss/_base.scss
```diff
@@ -93,8 +93,8 @@ main {
     & label {
       border: 1px solid transparent;
       display: block;
-      font-size: 18px;
-      font-weight: bold;
+      font-size: $medium-font;
```
muffinresearch: Good to start using vara for all of the font-sizes - helps keep things limited to specific sizes.

public/scss/_tooltip.scss
andymckay: alpha sorting css helps on that one too

public/scss/_tooltip.scss
```diff
@@ -6,7 +6,6 @@
   background: #F91200;
```
andymckay: we should make that a variable

andymckay:
r+, great to get this started earlier rather than later

public/scss/_tooltip.scss
muffinresearch: Good catch.

PR #156, opened 2015-07-03
Default text prop for Spinner component (fixes #154)

### LINKED ISSUE #154: Get rid of the string "Setting up"
Modify Spinner to make Loading the default.

muffinresearch:


public/jsx/views/card-details.jsx
```diff
@@ -57,7 +57,7 @@ module.exports = React.createClass({
         </div>
       );
     } else {
-      return <Spinner text={gettext('Setting up')}/>;
```
andymckay: Is there any difference to doing the gettext in the class as opposed to in the JSX? It looks like more code, but the result is same.

tests/test.spinner.jsx
```diff
@@ -0,0 +1,24 @@
...
+describe('Spinner', function() {
+
+  it('Uses loading as the default text', function() {
+    var Spinner_ = TestUtils.renderIntoDocument(<Spinner />);
+    var textNode = helpers.findByClass(Spinner_, 'text');
+    assert.equal(textNode.getDOMNode().firstChild.nodeValue, 'Loading');
+  });
+
+  it('Uses custom text as supplied', function() {
+    var Spinner_ = TestUtils.renderIntoDocument(<Spinner text="whatever" />);
```
andymckay: Ah and this test demonstrates the difference, you can pass a custom value.

andymckay:
r+

public/jsx/views/card-details.jsx
muffinresearch: The one in `getDefaultProps` is just a default which can be overridden by the component rendering Spinner as a prop.

PR #157, opened 2015-07-03
Add analytics (fixes #74)

### LINKED ISSUE #74: Set up Google Analytics
Analytics everywhere, let's create an client id and account for payments and add that in.

Then we'll need to start sprinkling events across the payments-ui like fairy dust.

muffinresearch:
* Tracking class is a ES6 class.
* There's not much in the way of instrumentation just yet more to come on that front.
* Adding this also moves us over to using the babel-loader in favour of the jsx-loader as React announced they're deprecating their own transpiler.

### DIFF
--- a/public/js/main.js
+++ b/public/js/main.js
@@ -2,10 +2,12 @@
 
 var $ = require('jquery');
 var App = require('app');
+var tracking = require('tracking');
 
 // Common ajax settings.
 $.ajaxSetup({
   dataType: 'json',
 });
 
+tracking.init();
 App.init();
--- a/public/js/tracking.js
+++ b/public/js/tracking.js
@@ -1,0 +1,87 @@
+'use strict';
+
+/* global ga */
+
+var settings = require('settings');
+
+
+class Tracking {
+
+  constructor() {
+    this.id = settings.tracking.id;
+    this.enabled = settings.tracking.enabled;
+    this.logPrefix = '[GA:' + (this.enabled ? 'ON' : 'OFF') + ']';
+    return this;
+  }
+
+  _ga() {
+    if (!this.initialized) {
+      throw new Error('Must call init() first');
+    } else {
+      ga.apply(window, Array.prototype.slice.call(arguments));
+    }
+  }
+
+  init() {
+    if (this.enabled === true) {
+      console.log('Tracking init (Analytics is ON)');
+      /*eslint-disable */
+      (function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+      (i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+      m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+      })(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+      ga('create', this.id, 'auto');
+      ga('send', 'pageview');
+      /*eslint-enable */
+    } else {
+      console.log('Tracking init. (Analytics is OFF)');
+    }
+    this.initialized = true;
+  }
+
+
+ /*
+  * Param           Type    Required  Description
+  * opts.category   String  Yes       Typically the object that
+  *                                   was interacted with (e.g. button)
+  * opts.action     String  Yes       The type of interaction (e.g. click)
+  * opts.label      String  No        Useful for categorizing events i
+  *                                   (e.g. nav buttons)
+  * opts.value      Number  No        Values must be non-negative.
+  *                                   Useful to pass counts (e.g. 4 times)
+  */
+  sendEvent(opts) {
+    if (!opts.category) {
+      throw new Error('sendEvent: opts.category is required');
+    }
+    if (!opts.action) {
+      throw new Error('sentEvent: opts.action is required');
+    }
+    if (this.enabled) {
+      this._ga('send', {
+        'hitType': 'event',
+        'eventCategory': opts.category,
+        'eventAction': opts.action,
+        'eventLabel': opts.label,
+        'eventValue': opts.value,
+      });
+    }
+    console.log(this.logPrefix, 'sendEvent', JSON.stringify(opts));
+  }
+
+ /*
+  * Should be called when a new view is show.
+  * Used in a SPA when the url is changed.
+  */
+  setPage(page) {
+    if (!page) {
+      throw new Error('setPage: page is required');
+    }
+    if (this.enabled) {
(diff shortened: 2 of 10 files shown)


public/js/tracking.js
```diff
@@ -0,0 +1,90 @@
+'use strict';
+
+/* global ga */
+
+var settings = require('settings');
+let instance = null;
+
+class Tracking {
```
kumar303: spiffy! Doesn't this file need to end in ``.jsx`` though? It seemed like the bable loader will not load js files but maybe I missed it.

tests/test.tracking.js
```diff
@@ -0,0 +1,88 @@
...
+var tracking = require('tracking');
+
+
+describe('tracking uninitialized', function() {
+
+  beforeEach(function() {
+    this.oldTrackingEnabled = tracking.enabled;
+    this.oldTrackingInitialized = tracking.initialized;
+    tracking.enabled = true;
+    tracking.initialized = false;
```
kumar303: When testing objects that have state I think it's pretty crucial to create a new instance to avoid leakage. Is there a specific reason that the Tracking object is a singleton? I can't see any reason. You could change the class so that it's not a singleton. Then you could start the test with a fresh instance like this:

````
var tracking = rewire('tracking');
var Tracking = tracking.__get__('Tracking');  // get the global Tracking class

beforeEach(function() {
  this.tracking = new Tracking();
});
````

File: public/jsx/views/complete-payment.jsx [hunk 3]
```diff
@@ -18,6 +19,10 @@ module.exports = React.createClass({
     userEmail: React.PropTypes.string.isRequired,
   },
 
+  componentDidMount: function() {
+    tracking.setPage('/payment/complete');
```
kumar303: does this have to be a URL? Could we do ``tracking.setPage('CompletePayment')`` ?

tests/test.tracking.js
```diff
@@ -0,0 +1,88 @@
...
+  });
+
+  afterEach(function() {
+    tracking.enabled = this.oldTrackingEnabled;
+    tracking.initialized = this.oldTrackingInitialized;
+  });
+
+  it('should throw when calling setPage', function() {
```
kumar303: Even though this is the first function, it should have ``tracking.initialized = false`` at the top in case someone re-orders the tests later

tests/test.tracking.js
```diff
@@ -0,0 +1,88 @@
...
+    sinon.stub(tracking, '_ga');
+  });
+
+  afterEach(function() {
+    tracking.enabled = this.oldTrackingEnabled;
+    tracking.initialized = this.oldTrackingInitialized;
+    tracking._ga.restore();
+  });
+
+  it('should throw if page not set', function() {
```
kumar303: this is already tested up above, right?

kumar303:
You'll need to rebase and fix up files from the `jsx` directory since those moved to `js`

kumar303:
So I guess tracking is disabled when running tests? I'm just thinking about tests such as the one for [CardListing](https://github.com/mozilla/payments-ui/blob/master/tests/test.card-listing.jsx#L22) which would trigger `componentDidMount()` and thus fire off tracking events. If that's the case you could disable it with rewire by doing `CardListing.__set__('tracking', sinon.stub())` but that might be annoying for every test.

public/js/main.js
```diff
@@ -2,10 +2,12 @@
...
 var $ = require('jquery');
 var App = require('app');
+var tracking = require('tracking');
 
 // Common ajax settings.
 $.ajaxSetup({
   dataType: 'json',
 });
 
+tracking.init();
```
kumar303: maybe call this after `App.init()` just so it doesn't block the UI at all.

public/jsx/views/complete-payment.jsx
kumar303: if nothing else, these should match the components/modules more closely. Maybe like: ``/complete-payment``, ``/card-listing``, ``/card-details``, and so on.

kumar303:
Shouldn't we be using Universal Analytics instead of Google Analytics? https://developers.google.com/analytics/devguides/collection/upgrade/guide#tagging-reference Maybe you are already, I'm not very familiar with either one.

kumar303:
r+wc

public/js/main.js
kumar303: but the more I think about that it's probably not possible.

tests/test.tracking.js
muffinresearch: No - this is checking an arg was sent.

tests/test.tracking.js
muffinresearch: That's what beforeEach etc is doing.

public/jsx/views/complete-payment.jsx
muffinresearch: Yeah - I'll check the docs. I don't see why it can't be an arbitrary string.

public/js/tracking.js
muffinresearch: I've made js and jsx both be run through babel see https://github.com/mozilla/payments-ui/pull/157/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR18

tests/test.tracking.js
muffinresearch: So the thought behind this was initially about avoiding the setup piece on every require. A module exporting an instance should implicitly be a singleton (afaict though I've seen some things saying you might get a different object if your path for a require is different) , I just made this more explicit in the code.

The reason for the single instance was to track state for when the setup had been completed. This way we can show errors if setPage / sendEvent gets called but the analytics code hasn't been inited.

muffinresearch:
> Shouldn't we be using Universal Analytics instead of Google Analytics? https://developers.google.com/analytics/devguides/collection/upgrade/guide#tagging-reference Maybe you are already, I'm not very familiar with either one.

This is universal. The way to tell is that this is pointing at analytics.js and not ga.js.

public/jsx/views/complete-payment.jsx
muffinresearch: Hmm according to the docs it must start with a '/' https://developers.google.com/analytics/devguides/collection/analyticsjs/pages

tests/test.tracking.js
kumar303: I know it's not as sexy but this would probably be simpler as a module object since that's how you're using it:
````
var initialized = false;

module.exports = {
  init: () => {
    // ...
    initialized = true;
  },

  sendEvent: opts => {
    // ...
  },
};
````
I still think it would be easier to test in isolation by making a new instance of the class for each test though. This way you don't have to tear down or set up anything.

PR #160, opened 2015-07-03
initial pass at HTML emails

andymckay:


andymckay:
for some value of "ready"

public/scss/_email.scss
```diff
@@ -0,0 +1,34 @@
+#email {
+    background-color: #F2F2F2;
+    padding-top: 1em;
+
+    #wrapper {
+        background-color: white;
```
muffinresearch: Should use #fff

public/scss/_email.scss
```diff
@@ -0,0 +1,34 @@
...
+        border: 1px solid #DDDDDD;
```
muffinresearch: #ddd

public/scss/_forms.scss
```diff
@@ -144,16 +144,18 @@ label:focus ~ .input-help-focused {
...
 
-.card-icon {
-    opacity: 0;
-    position: absolute;
-    right: 8px;
-    top: 50%;
-    margin-top: -13px;
-    z-index: 10;
-}
+form {
```
muffinresearch: We won't want to do this as this will mean the card icon stuff will only work inside a form.

andymckay: Thats only the position of the card icon within its parent I believe? In the case of the email it shoves it into the bottom right hand corner miles away from where it should be.

muffinresearch: Is that because you don't have `position: relative` on the parent?

muffinresearch:
I feel like this should probably be a different target (e.g. separate file in the grunt task output). This way it can use all the existing styles for things like fonts + buttons but it won't add to the static file weight.

andymckay:
So then you start up grunt with the seperate target before we run the premailer script?

public/scss/_forms.scss
andymckay: All I really want in card-icon is this: https://github.com/mozilla/payments-ui/blob/0279c768a05411e6cdb0b8e3640436c5e732f292/public/scss/_cardicons-sprite.scss#L1, not all the rest.

muffinresearch:
> So then you start up grunt with the seperate target before we run the premailer script?

As you're grabbing this from what is already generated you could just have a different file built e.g. /dist/email.css - this would get built in the same way the public file are for the docker build. If you were developing this locally then yes at that point you'd need to run something like `grunt sass:email` to build just the css for the emails.

public/scss/_email.scss
```diff
@@ -0,0 +1,34 @@
...
+            white-space: nowrap;
+        }
+
+        td {
+            text-align: right;
+            padding-left: 1em;
+        }
+
+        .card-icon {
+            background: url(/img/cardicons-sprite.svg) no-repeat;
```
muffinresearch: This is already defined - you don't need to redefine it.

muffinresearch:
If you create a new target another benefit is you can leave the forms stuff out completely negating the need to change the specificity of those selectors.

Just copy https://github.com/mozilla/payments-ui/blob/master/public/scss/main.scss to email.scss
Remove everything you don't need probably something like this: 

```
@import 'lib/normalize';

@import 'inc/vars';
@import 'inc/mixins';

@import '_typography';
@import '_base';
@import '_buttons';
@import '_cardicons-sprite';

// ADD The rest of your email rules HERE or you could use a partial e.g. `@import '_email'` or similar.

```

Then add a new target here: https://github.com/mozilla/payments-ui/blob/master/tasks/sass.js

e.g: 

```
  email: {
    options: {
      outputStyle: 'expanded'
    },
    files: {
      'public/dist/email.css': 'public/scss/email.scss',
    },
  },
```

The sass task is run automatically for the docker-hub builds. (running `grunt sass`) will build all the targets including this new one.

If you need to build it yourself just run `grunt sass:email`

Then update your paths in the scripts to point at `/dist/email.css` and you should be good.

andymckay:
@muffinresearch fixed up as per the comments. The pre-mailed HTML isn't great, there's some issues I will file bugs for so we can work on those together. But its something at least.

andymckay:
Filed mozilla/payments-ui#175 and mozilla/payments-service#92.

muffinresearch:
You're still missing public/scss/email.scss in the current diff.

andymckay:
Missing file added.

muffinresearch:
Rawk! :shipit: r+

PR #169, opened 2015-07-06
Fix linting in nested directories

kumar303:
This diff will be easier to read once https://github.com/mozilla/payments-ui/pull/161 lands


muffinresearch:
Good catch r+

Issue #175, opened 2015-07-07
Card icon images in emails


andymckay-limited-access:
The CSS has card-icon being a relative link, e.g.: ../img/cardicons-sprite.svg

when the premailer runs, it doesn't turn this into an absolute url, so that becomes a 404 in the email.

muffinresearch:
We need a mixin for this.

muffinresearch:
This will be fixed by #175.

PR #177, opened 2015-07-08
Make image and font-paths configurable (fixes #176)

### LINKED ISSUE #176: Make font and image paths configurable

muffinresearch:
Local paths need to be relative otherwise moving the styles around (e.g. for the styleguide) will break.

Using `DEV=1 grunt sass` will put the css image/font urls from '/' which should work better for the email case if my understanding on #175 is correct.


muffinresearch:
Going land this as I need to use it. Feel free to post-review.

PR #182, opened 2015-07-09
Add initial management page (fixes #171)

### LINKED ISSUE #171: Make entry points for separate modules for management and payments

muffinresearch:
management

Notes:
* Webpack conf now generates a commons.js file which is the js shared between the payments and management interface.
* Styles are just a starting point - will defintely need updating to spec in due course.
* Commons.js needs minification conf (I'll file a bug for that)


public/js/apps/management/app.jsx
```diff
@@ -0,0 +1,86 @@
...
+var Management = React.createClass({
+
+  displayName: 'ManagementApp',
+
+  render: function() {
+    return (
+
+      <div>
+        <header className="top-nav">
+          <h1 className="logo">Firefox Payments</h1>
```
kumar303: hmm, should this be localized? I'm not sure

public/js/apps/management/app.jsx
```diff
@@ -0,0 +1,86 @@
...
+          <button>{gettext('Sign Out')}</button>
+        </header>
+
+        <div className="content">
+          <div className="user">
+            <p>Hello, placeholder@placeholder.com</p>
```
kumar303: this needs localization. We could fix it up later when we add a real email though.

public/js/apps/management/main.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+var $ = require('jquery');
+var App = require('./app');
+
+// Common ajax settings.
+$.ajaxSetup({
+  dataType: 'json',
```
kumar303: hmm, it's possible that the [header config](https://github.com/mozilla/payments-ui/blob/master/public/js/actions/user.jsx#L25) might erase this setting.

public/js/apps/payment/main.js
```diff
@@ -1,8 +1,8 @@
 'use strict';
 
 var $ = require('jquery');
-var App = require('app');
 var tracking = require('tracking');
+var App = require('./app');
```
kumar303: nice. I feel like we should do relative imports more often

public/js/components/card-icon.jsx
```diff
@@ -1,6 +1,6 @@
 'use strict';
 
-var classNames = require('classnames');
+var cx = require('classnames');
```
kumar303: why cx?

public/js/views/management.jsx
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
```
kumar303: what is this file used for? It seems like ``management/app.jsx`` replaced it, no?

kumar303:
r+wc

public/js/apps/management/app.jsx
muffinresearch: Yeah, that's what I was thinking too given there's no format func yet.

public/js/components/card-icon.jsx
muffinresearch: It's a convention in react stuff.

public/js/views/management.jsx
muffinresearch: Oh yeah - this should be deleted.

public/js/apps/management/app.jsx
kumar303: oh bummer, I thought gettext already had formatting.

public/js/components/card-icon.jsx
kumar303: wfm

public/js/apps/management/main.js
muffinresearch: Yep looks like it. I'll file a bug.

PR #190, opened 2015-07-13
more red text

andymckay:
* As per issue: https://github.com/mozilla/payments/issues/70

Before:

screenshot 2015-07-13 10 55 55

After:

screenshot 2015-07-13 10 55 29

### DIFF
--- a/public/scss/_base.scss
+++ b/public/scss/_base.scss
@@ -96,3 +96,10 @@ main {
       text-align: right;
   }
 }
+
+.app-error {
+    @include header-font();
+    color: $error-text-color;
+    margin-top: 50%;
+    text-align: center;
+}


muffinresearch:
This should be it its own file ideally so we align css to modules/components more closely.

public/scss/_base.scss
```diff
@@ -96,3 +96,10 @@ main {
       text-align: right;
   }
 }
+
+.app-error {
+    @include header-font();
+    color: $error-text-color;
+    margin-top: 50%;
```
muffinresearch: This won't quite make it vertically centered.

muffinresearch:
I'll land this as is, it's definitely better than what we had before and we can easily iterate.

PR #194, opened 2015-07-14
Move app-error styling (fixes #193)

### LINKED ISSUE #193: Move error styling

muffinresearch:
This still needs to have the app error component updated to do an overlay as it's not currently implemented in this way afaict. But this at least means the styles are ready for that.

Added to the styleguide to demonstrate what it looks like with something underneath.

payments_ui_styleguide___app_error


andymckay:
Having it in an overlay like that is awesome, thanks for doing that. Also, got to play with it in the styleguide for testing. Woot.

r+

kumar303:
r+. We really need to dig into all error cases and make sure they are routed to the error handler (or a local handler for form errors).

andymckay:
Let's file bugs on all the error cases we aren't catching, e.g. repeated subscription.

Issue #197, opened 2015-07-14
Write payments-ui translations to payments-l10n


andymckay-limited-access:
So they can be translated as per https://bugzilla.mozilla.org/show_bug.cgi?id=1173423 supports mozilla/payments#87

PR #198, opened 2015-07-14
push pot and po generation into payments-l10n

andymckay:


andymckay:
connects to #197

tasks/abideExtract.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
      src: [
        'public/dist/*.js',
      ],
-     dest: 'locale/templates/LC_MESSAGES/messages.pot',
+     dest: '../payments-l10n/locale/templates/LC_MESSAGES/payments-ui.pot',
```
kumar303: what does `../payments-l10n` point to? Does that need to be a separate source checkout? If so, the dev docs need updating: https://github.com/mozilla/payments-ui#developers

kumar303:
r+wc

PR #200, opened 2015-07-16
Add hot module reloading (fixes #199)

### LINKED ISSUE #199: Add hot module loading support

muffinresearch:
If you want to try it and it doesn't work, make sure you do have the latest grunt-webpack installed as per the package.json.


README.md
```diff
@@ -48,6 +48,17 @@ are identical and can't get mangled by a broken deps installation.
 
 We're using webpack to build a JS bundle.
 
+### Hot module reloading
+
+If you run the webpack-dev-server you can get hot module reloading.
+
+For example run `grunt serve`
+
+And then visit http://localhost:8080/webpack-dev-server/management.html
```
kumar303: how hard is it to get it working on `pay.dev`? The main reason is that we need to be on the same domain as the API.

karma.shared.js
```diff
@@ -2,16 +2,19 @@
 
 /* This is the shared karma config */
 
+var merge = require('lodash.merge');
 var RewirePlugin = require('rewire-webpack');
-var webpackConfig = require('./webpack.config.js');
+var webpackConfig = require('./webpack.config');
 
+
+var newWebpackConfig = merge({}, webpackConfig);
```
kumar303: this works but we also have the [object-assign](https://github.com/mozilla/payments-ui/blob/master/package.json#L16) polyfill which is used elsewhere like `assign({}, ...)`

kumar303:
Nice! r+wc. I didn't try it out yet. It would be really great if we could do this on `pay.dev` (with a setting or whatever) but this is good start.

karma.shared.js
kumar303: Thinking more about this, I'd prefer `object-assign` just so we don't add a new dep (even though it's a dev dep).

README.md
muffinresearch: Hmm, yeah - will need to think about that. It'll probably need Charles Proxy or maybe we can do something with nginx?

karma.shared.js
muffinresearch: Ah ok, I didn't know about that polyfill. I'll ditch the lodash dep.

muffinresearch: @kumar303 actually assign isn't going to work for this. See http://jsbin.com/nuhawi/edit?js,console for why.

README.md
kumar303: As discussed, do the simplest thing first and land it :) Looking at the CORS cookie thing though, I think we can run the dev server on the host and talk to `pay.dev/api/` with just a few setting changes in the service.

karma.shared.js
kumar303: Also as discussed, carry on! Maybe we could just remove assign everywhere for now? A quick grep suggests to me that we aren't even using it, at least not in prod code.

kumar303:
ah, I added some comments thinking you hadn't merged this yet. You can ignore them. Glad this landed and is moving forward.

PR #206, opened 2015-07-17
Add modal component (fixes #188)

### LINKED ISSUE #188: Implement Modal in React

muffinresearch:
The initial component for a modal. Likely to be tweaked as it's incorporated via redux.

payments_ui_styleguide___modal_dialogue

### DIFF
--- a/public/js/components/modal.jsx
+++ b/public/js/components/modal.jsx
@@ -1,0 +1,46 @@
+'use strict';
+
+var React = require('react');
+var gettext = require('utils').gettext;
+var cx = require('classnames');
+
+
+module.exports = React.createClass({
+
+  displayName: 'Modal',
+
+  propTypes: {
+    children: React.PropTypes.array.required,
+    close: React.PropTypes.func,
+    showModal: React.PropTypes.bool,
+    title: React.PropTypes.string,
+  },
+
+  getDefaultProps: function() {
+    return {
+      showModal: true,
+    };
+  },
+
+  render: function() {
+
+    var classes = cx(['modal', {'active': this.props.showModal}]);
+
+    return (
+      <div className={classes} onClick={this.props.close}>
+        <div className="inner">
+          <header>
+            {this.props.title ? <h2>{this.props.title}</h2> : null}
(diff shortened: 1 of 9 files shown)


public/js/components/modal.jsx
```diff
@@ -0,0 +1,46 @@
...
+var gettext = require('utils').gettext;
+var cx = require('classnames');
+
+
+module.exports = React.createClass({
+
+  displayName: 'Modal',
+
+  propTypes: {
+    children: React.PropTypes.array.required,
```
kumar303: This should be `isRequired`. Hmm, too bad these create warnings not errors. If there is a flag in react to throw errors we should try it.

kumar303: bah, this might not be possible http://stackoverflow.com/questions/27006782/force-reactjs-to-throw-real-errors-when-proptypes-validation-fails

public/js/components/modal.jsx
```diff
@@ -0,0 +1,46 @@
...
+    close: React.PropTypes.func,
```
kumar303: from usage, this looks like it should be required

public/js/components/modal.jsx
```diff
@@ -0,0 +1,46 @@
...
+    showModal: React.PropTypes.bool,
+    title: React.PropTypes.string,
```
kumar303: if this isn't required maybe set a default so it's not blank?

kumar303:
I kinda feel like this should have a quick test, at least one that passes a sinon stub to the close function and makes sure it gets triggered. It's not a complex component but adding one test would also alert us in case of any runtime errors.

kumar303:
r+wc

public/js/components/modal.jsx
muffinresearch: Yeah, this is fixed in my new branch. It should also be node not array or object.

muffinresearch:
>  I kinda feel like this should have a quick test, at least one that passes a sinon stub to the close function and makes sure it gets triggered. It's not a complex component but adding one test would also alert us in case of any runtime errors.

Tests are going to be added in due course.

PR #208, opened 2015-07-17
Step back to babel stage 2 (fixes #207)

### LINKED ISSUE #207: Step back to babel Stage 2 and enable just es7.objectRestSpread from stage 1
To prevent us using features of future ES versions that may never see prime time we should look at stepping back grom using stage 1 and try enabling just the object rest spread operator from ES7 which is needed for JSX.

muffinresearch:
Rather than use all of stage 1 let's just use the one bit that makes JSX work `es7.objectRestSpread`

See https://babeljs.io/docs/usage/experimental/

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -21,9 +21,9 @@ module.exports = {
       {
         exclude: /(node_modules|bower_components)/,
         test: /\.jsx?$/,
-        // Stage 1 to enable ES7 rest properties.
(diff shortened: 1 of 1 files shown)


kumar303:
canoe-grey-squirrel-taxidermy-mount-_12576-for-sale-_-the-taxidermy-store

r+

PR #225, opened 2015-07-23
User sign in/out for management (fixes #201 fixes #192)

### LINKED ISSUE #192: Sign out
Link up the sign out page to a method that signs a user out.

Supports mozilla/payments#85

### LINKED ISSUE #201: Sign in to management app
By using the [sign-in](http://payments-service.readthedocs.org/en/latest/topics/api.html#sign-in) endpoint from https://github.com/mozilla/payments-service/issues/94, allow user to sign in with an FxA username/password. This will need to share the [same client ID](https://github.com/mozilla/payments-service/blob/master/payments_service/settings/base.py#L224) as payments-service.

kumar303:
Notes:
* ~~I forgot to make a separate commit for the fxa-relier-client dist files, sorry about that~~
* The header sign-in / sign-out button works but all of that might be temporary. I'm not sure if that's the right thing - we probably want to do an auto sign-in when the page loads. That can come later.
* I didn't add tests for the sign-in buttons because of that previous point.


public/js/actions/user.js
```diff
@@ -2,12 +2,24 @@
...
+      'X-CSRFToken': csrfToken,
+    },
+  });
+}
 
 
-export function signIn(accessToken) {
-  return function(dispatch) {
+export function tokenSignIn(accessToken) {
+  return dispatch => {
```
muffinresearch: I know the params are optional for arrow funcs with one arg, but I feel like it's more readable with them. It seems weird to me that you can have
```javascript
() = > {};
foo => {}; // <- the outlier
(foo) => {};
(foo, bar) => {};
```

public/js/views/management.jsx
```diff
@@ -58,7 +86,7 @@ module.exports = React.createClass({
                   <a className="button"
                      href="https://mozilla.org/"
                      target="_blank">{gettext('Change')}</a>
-                  <p>placeholder@placeholder.com</p>
+                  <p>{this.props.user ? this.props.user.email : ''}</p>
```
muffinresearch: Could this not output anything if this.props.user is falsey e.g something like:
```
{this.props.user ? <p>{this.props.user.email}</p> : null} 
```

muffinresearch:
As you fix-up the conflicts if you can update any relevant bits to es6 this looks good to me.

r+wc+ES6

public/js/actions/user.js
kumar303: It's less typing though. Embrace the future! But yeah I see what you mean. Maybe we should give a try and see how bad the readability is? I think new language syntax sometimes takes getting used to. I once saw a person new to python write code like this:
````python
def foo():
    if the_thing:
        do_stuff()
    # END if
# END foo()
````
because they didn't think the indentation was readable enough for block endings.

muffinresearch: Yeah, let's see. For me it's more about consistency than anything else. But yes - maybe I get used to it soon enough. It's a brave new world!

back-to-the-future

PR #230, opened 2015-07-24
Sign into management with a shared token (fixes #202)

### LINKED ISSUE #202: Sign in to management app with shared token
In addition to supporting a full FxA sign-in (#201) we should support access token sign-in similar to the purchase flow. The story goes like this:

* user logs into third party (Mozilla Concrete)
* user does not make a purchase but instead clicks 'manage my payment account'
* Mozilla Concrete links to the management app passing the scoped access token in the query string

kumar303:


kumar303:
Example app for this: https://github.com/mozilla/payments-example/pull/23

public/js/apps/management/app.jsx
```diff
@@ -1,20 +1,33 @@
...
 
 import reduxConfig from 'redux-config';
 import * as managementActions from 'actions/management';
 import * as userActions from 'actions/user';
+import { parseQuery } from 'utils';
 
 import ModalError from 'views/modal-error';
-import Management from 'views/management';
-import ManageCards from 'views/manage-cards';
+import { default as DefaultManagement } from 'views/management';
```
muffinresearch: I think you could just do `DefaultManagement from blah` but this is probably nicer as it's more explicit.

public/js/apps/management/app.jsx
```diff
@@ -1,20 +1,33 @@
...
+import { default as DefaultManageCards } from 'views/manage-cards';
 
 
 export default class ManagementApp extends Component {
 
+  static propTypes = {
+    ManageCards: PropTypes.object,
```
muffinresearch: I didn't get around to checking if this should be node, but I suspect not since this is a class not a function returned by createClass.

public/js/apps/management/app.jsx
```diff
@@ -59,9 +78,7 @@ export default class ManagementApp extends Component {
 export function init() {
   React.render((
     <Provider redux={reduxConfig}>
-      {function() {
-        return <ManagementApp/>;
-      }}
+      {() => <ManagementApp/>}
```
muffinresearch: Nice and compact, I like it!

tests/helpers.jsx
```diff
@@ -67,6 +67,17 @@ export const cvvError = {
 };
 
 
+export function findById(component, id){
```
muffinresearch: Handy.

tests/test.management.jsx
```diff
@@ -0,0 +1,61 @@
...
+
+    TestUtils.Simulate.click(button.getDOMNode());
+    assert.equal(fakeUserSignIn.called, true);
+  });
+
+  it('should sign out on button click', function() {
+    var view = mountView({user: {signedIn: true}});
+    var button = helpers.findById(view, 'sign-in-toggle');
+
+    TestUtils.Simulate.click(button.getDOMNode());
```
muffinresearch: Does TestUtils magically add `getDOMNode()`? If so I probably factored out more of those than I needed to before.

muffinresearch:
r+wc

tests/test.management.jsx
muffinresearch: Actually the docs say it's deprecated so we should probably continue along the lines of removing all usage of getDOMNode() see https://facebook.github.io/react/docs/component-api.html#getdomnode

public/js/apps/management/app.jsx
kumar303: aha, it looks like we can use `PropTypes.element` https://facebook.github.io/react/docs/reusable-components.html

tests/test.management.jsx
kumar303: oh huh, yeah, `click(button)` worked just fine.

public/js/apps/management/app.jsx
muffinresearch: Ah that's good to know.

PR #231, opened 2015-07-27
Remove all traces of rewire packages

muffinresearch:


muffinresearch:
Minimal clean-up - @kumar303 feel free to r? post landing.

public/.eslintrc
```diff
@@ -1,7 +1,7 @@
 {
   "env": {
     "browser": true,
-    "es6": "true",
+    "es6": true,
```
kumar303: heh

karma.shared.js
```diff
@@ -1,25 +1,19 @@
...
 // Remove the bits from the shared config
 // that we don't want for tests.
 delete newWebpackConfig.output;
 delete newWebpackConfig.entry;
 
-// Add this just for testing:
-newWebpackConfig.plugins = [
-  new RewirePlugin(),
-];
+newWebpackConfig.plugins = [];
```
kumar303: this line can just be deleted

kumar303:
post merge r+ :) looks good

karma.shared.js
muffinresearch: I don't think it can because we don't want commons here. See https://github.com/mozilla/payments-ui/blob/master/webpack.config.js#L31

kumar303: ah, right

PR #238, opened 2015-07-28
Add no cards message (fixes #237)

muffinresearch:
* Renamed ManageCards/manage-cards to PayMethods/pay-methods
* Added tests for PayMethods container.
* Added no-cards message.

PR #240, opened 2015-07-28
Split out token (fixes #235)

### LINKED ISSUE #235: Move braintree token request out of CardDetails
This should happen in redux not `componentWillMount`.

muffinresearch:
This is based on #238 so for the purposes of review ignore the first commit. I'll rebase on master once #238 lands.


public/js/actions/user.js
```diff
@@ -127,3 +127,25 @@ export function userSignOut(jquery=$) {
...
+export function getBraintreeToken() {
+  console.log('Requesting braintree token');
+  return dispatch => {
+    $.ajax({
+      method: 'post',
+      url: '/api/braintree/token/generate/',
+      context: this,
+    }).then(data => {
+      dispatch({
+        type: actionTypes.RX_BRAINTREE_TOKEN,
```
muffinresearch: Happy to have ideas on a better name for this - but this seemed to be concise and said what it does.

kumar303: What does `RX` mean? I've been trying to make action type names follow the action which would make this `GET_BRAINTREE_TOKEN`, only one more char.

public/js/actions/user.js
```diff
@@ -127,3 +127,25 @@ export function userSignOut(jquery=$) {
...
+        user: {
+          braintreeToken: data.token,
+        },
```
kumar303: since this action only returns a token, I don't think you need to wrap it in a user object. Could just be:
````
dispatch({
  type: ...,
  braintreeToken: data.token,
})
````

Comment (kumar303) [inline public/js/actions/user.js]: we generally keep data keys lower case with underscores. I guess that's a Python/API thing but it might be nice for consistency to call this `braintree_token`.

File: public/js/apps/payment/app.jsx [hunk 3]
```diff
@@ -67,7 +68,10 @@ export default class PaymentApp extends Component {
             } else {
               console.log('rendering purchase flow');
               return (
-                <Purchase user={result.user} productId={state.productId} />
+                <Purchase
+                 productId={state.productId}
```
kumar303: is this indented for vertical alignment? Everywhere else we've been doing two space indents.

public/js/reducers/user.js
```diff
@@ -4,19 +4,19 @@ const defaults = {
   signedIn: false,
   email: null,
   payment_methods: [],
+  braintreeToken: null,
 };
 
 export default function user(state, action) {
+  console.log('user store: got action', action);
```
kumar303: the reason I haven't been logging at the top level is that a reducer gets called for every action. In other words, this will be noisy because it will include ignored actions. With that said, adding a log statement to every if statement is also maybe a little overkill. It has helped me in the past though.

public/js/reducers/user.js
```diff
@@ -24,5 +24,13 @@ export default function user(state, action) {
     });
   }
 
+  if (action.type === actionTypes.RX_BRAINTREE_TOKEN) {
+    // As we want to stay signed in if we are
+    // this reducer inherits state.
```
kumar303: this sentence is a bit awkward.

public/js/reducers/user.js
```diff
@@ -24,5 +24,13 @@ export default function user(state, action) {
...
+    return Object.assign({}, defaults, state, {
+      braintreeToken: action.user.braintreeToken,
```
kumar303: per previous comments, I think it's nice to keep this data specific to the action and just call it `action.braintree_token`

public/js/views/card-details.jsx
```diff
@@ -11,58 +9,25 @@ import { default as tracking } from 'tracking';
...
-  constructor(props) {
-    super(props);
-    this.state = {
-      braintree_token: false,
-    };
-  }
-
   componentDidMount() {
-    console.log('Requesting braintree token');
-    // TODO: move this to a purchase action.
```
kumar303: woohoo!

public/js/views/purchase.jsx
```diff
@@ -55,11 +61,23 @@ export default class Purchase extends Component {
...
+                {...bindActionCreators(userActions, result.dispatch) }
+              />
+            );
           } else {
             console.log('rendering card entry');
-            return <CardDetails productId={props.productId} />;
+            return (
+              <CardDetails
+               braintreeToken={result.user.braintreeToken}
+               productId={props.productId}
```
kumar303: two space indent?

public/js/views/card-details.jsx
```diff
@@ -11,58 +9,25 @@ import { default as tracking } from 'tracking';
 export default class CardDetailsView extends Component {
```
kumar303: nit: could we call this `CardDetails` to match the file?

public/js/views/purchase.jsx
```diff
@@ -55,11 +61,23 @@ export default class Purchase extends Component {
                 {...bindActionCreators(purchaseActions, result.dispatch)}
               />
             );
+          } else if (!result.user.braintreeToken) {
+            console.log('Retreiving Braintree Token');
+            return (
+              <BraintreeToken
+                {...bindActionCreators(userActions, result.dispatch) }
+              />
```
kumar303: I hummed and hawed about whether this needed to be a separate component but I decided I liked it :) What I really like is that the component with the connector (Purchase) becomes the only render() function with if statements in it. We should strive to only put if statements in connectors. This makes it so tests of our connectors will cover all integration of smaller units.

public/js/views/card-details.jsx
```diff
@@ -11,58 +9,25 @@ import { default as tracking } from 'tracking';
...
-        </div>
-      );
-    } else {
-      return <Spinner />;
-    }
+    return (
+      <div className="card-details">
+        <ProductDetail productId={this.props.productId} />
+        <CardForm
+          data-token={this.props.braintreeToken}
```
kumar303: I think `data-token` is pre-react cruft. While you're in here, you might as well rename it to `braintreeToken={...}` because the CardForm code will read a lot cleaner.

tests/test.data-store.js
```diff
@@ -90,6 +93,7 @@ describe('user', function() {
     });
 
     assert.deepEqual(user, {
+      braintreeToken: null,
       signedIn: false,
       email: null,
       payment_methods: [],
```
kumar303: You should add two tests here:
1. receive a braintree token action and make sure it's added to the user object
2. pre-populate a user object, receive a braintree token action, and make sure the old user object was preserved

tests/test.payment-app.jsx
```diff
@@ -68,6 +68,7 @@ describe('Payment App', function() {
 
   it('should render a purchase page', function() {
     var user = {
```
kumar303: I think it's time to should start doing this instead:
````
import { defaultUser } from 'actions/user';
...
var user = Object.assign({}, defaultUser, {email: 'f@f.com', signedIn: true});
````

There are a couple tests now that can benefit from that.

File: tests/test.purchase.jsx [hunk 14]
```diff
@@ -68,6 +73,14 @@ describe('Purchase', function() {
 
   it('should render a card entry form', function() {
     var View = mountView();
+
+    store.dispatch({
+      type: actionTypes.RX_BRAINTREE_TOKEN,
+      user: Object.assign({}, defaultUser, {
+        braintreeToken: 'braintree-token',
+      }),
```
kumar303: this isn't actually what the token action does. It only dispatches a token, not previous user data.

public/js/views/purchase.jsx
```diff
@@ -4,18 +4,21 @@ import { bindActionCreators } from 'redux';
...
 export default class Purchase extends Component {
 
   static propTypes = {
-    CardDetails: PropTypes.object,
-    CardListing: PropTypes.object,
-    CompletePayment: PropTypes.object,
+    CardDetails: PropTypes.func,
+    CardListing: PropTypes.func,
+    CompletePayment: PropTypes.func,
+    braintreeToken: PropTypes.string,
```
kumar303: I think you meant `BraintreeToken`

tests/test.purchase.jsx
```diff
@@ -97,6 +110,13 @@ describe('Purchase', function() {
...
     });
 
+    store.dispatch({
+      type: actionTypes.RX_BRAINTREE_TOKEN,
+      user: Object.assign({}, defaultUser, {
+        braintreeToken: 'braintree-token',
+      }),
+    });
+
     store.dispatch(purchaseActions.payWithNewCard());
```
kumar303: if you add a quick test here that makes sure the BraintreeToken component is rendered when `user.braintreeToken` is null then we'll have coverage of an important part of the payflow logic.

kumar303:
I think it needs a few more tests as noted.

r+wc

public/js/actions/user.js
kumar303: I thought more about this and since it's the *result* of the action it doesn't make perfect sense to follow the action name in the same verb tense. That made me realize that `RX` stands for receive :) However, I think you could use `GOT_BRAINTREE_TOKEN` and get the same effect with less cryptic abbreviation.

public/js/actions/user.js
muffinresearch: I saw that, however, since I'm setting the name on the JS side it seems reasonable to use camelCase :camel:

public/js/apps/payment/app.jsx
muffinresearch: Yeah I've seen both. Let's standardize on 2.

public/js/reducers/user.js
muffinresearch: Yeah this was what I was asking about yesterday. I'm going to leave this for now as it's consistent with everything else we've been doing. It makes it very clear where things have come from. Let's discuss this and if we still want to change it we can do so in a follow-up.

tests/test.purchase.jsx
muffinresearch: oops, good catch.

public/js/views/card-details.jsx
muffinresearch: Good idea. We didn't need to render this as an attr any more either.

tests/test.payment-app.jsx
muffinresearch: ooh nice idea!

PR #245, opened 2015-07-29
List subscriptions in management app (fixes #233)

### LINKED ISSUE #233: List subscriptions in management interface
Show a list of the user's subscriptions. See [UX mock](http://payments.readthedocs.org/en/latest/_images/ux-management-wireframe.png) for details but it's basically this:

screenshot 2015-07-27 11 40 02

Supports https://github.com/mozilla/payments/issues/85

kumar303:


public/scss/_subscriptions.scss
```diff
@@ -0,0 +1,56 @@
...
+            float: left;
+        }
+        .change-pay-account {
+            float: right;
+        }
+    }
+
+    .clear {
+        clear: both;
+    }
```
kumar303: I swear I tried `.subnav:after` but it wasn't working. If you can toss in a specific code block I can try it. Knowing css I probably had something wrong but since there are no errors in css (only tears) I'll never know.

public/js/views/management.jsx
```diff
@@ -26,6 +29,11 @@ export default class Management extends Component {
     }
   }
 
+  showSubscriptions = click => {
+    click.preventDefault();
```
muffinresearch: `click` seems a bit unconventional since it's an event object not a click per se.

public/js/apps/management/app.jsx
```diff
@@ -56,7 +59,10 @@ export default class ManagementApp extends Component {
       ));
     }
 
-    children.push(<Management {...boundMgmtActions} {...boundUserActions}
+    children.push(<Management {...boundMgmtActions}
+                              {...boundUserActions}
+                              {...boundSubscriptionActions}
```
muffinresearch: Hah, I'm adding one here too - I wonder if we can merge the actions into one. `bindActionCreators` could do with taking a list.

muffinresearch:
This PR means you win the job of moving this to an action https://github.com/mozilla/payments-ui/blob/master/public/js/components/card-form.jsx#L129 :+1:

public/js/actions/subscriptions.js
```diff
@@ -0,0 +1,33 @@
...
+
+import * as actionTypes from 'constants/action-types';
+
+import * as appActions from './app';
+
+
+export function getUserSubscriptions(jquery=$) {
+  return dispatch => {
+    dispatch({
+      type: actionTypes.LOADING_USER_SUBSCRIPTIONS,
```
muffinresearch: This is a different convention to the one where we're using a component that calls the action and just returns a spinner. Should we consolidate that (either way)?

public/js/reducers/user.js
```diff
@@ -33,5 +34,19 @@ export default function user(state, action) {
     });
   }
 
+  if (action.type === actionTypes.LOADING_USER_SUBSCRIPTIONS) {
+    console.log('user: initializing from', action);
+    return Object.assign({}, defaults, state, {
+      subscriptions: defaults.subscriptions,
```
muffinresearch: Elsewhere I've been calling this defaultState just to make it a bit more distinct from `default`

public/js/components/subscription-list.jsx
```diff
@@ -0,0 +1,45 @@
...
+      });
+
+      if (subs.length) {
+        return (
+          <ul className="subscription-list">
+            {subs}
+          </ul>
+        );
+      } else {
+        return <p>{gettext('You haven\'t subscribed to anything yet.')}</p>;
```
muffinresearch: You can use double quotes here. Eslint shouldn't complain any more.

public/js/components/subscription-list.jsx
```diff
@@ -0,0 +1,45 @@
...
+    console.log('subscriptions:', this.props.subscriptions);
+
+    if (this.props.subscriptions === null) {
+      return <Spinner/>;
+    } else {
+
+      var subs = [];
+      this.props.subscriptions.forEach((data) => {
+        subs.push(
+          <li>
```
muffinresearch: React is probably telling you that you should have a key on this.

public/js/components/subscription.jsx
```diff
@@ -0,0 +1,42 @@
...
+    }).isRequired,
+  }
+
+  render() {
+    var productData = products.get(this.props.seller_product.public_id);
+
+    return (
+      <div className="subscription">
+        <div className="product">
+          <img src={productData.img}/>
```
muffinresearch: Needs an alt attribute. If it doesn't add any value (e.g. it's a dupe of the heading) make it an empty string.

public/js/components/subscription.jsx
```diff
@@ -0,0 +1,42 @@
...
+          <h2 className="seller">{productData.seller.name.en}</h2>
+          <div className="description">{productData.description.en}</div>
```
muffinresearch: p might be a better fit than a div here.

public/js/components/subscription.jsx
```diff
@@ -0,0 +1,42 @@
...
+        </div>
+        <div className="price">
+          <div>{productData.price.en}</div>
+          <div>/ {gettext('month')}</div>
+        </div>
+        <div className="clear"></div>
```
muffinresearch: http://yakshavings.co.uk/img/nope.gif

public/js/products.js
```diff
@@ -1,4 +1,14 @@
...
   'mozilla-concrete-brick': require('json!mozilla-concrete-brick'),
   'mozilla-concrete-mortar': require('json!mozilla-concrete-mortar'),
 };
+
+export default productData;
+
+
+export function get(id) {
+  if (!productData[id]) {
+    throw new Error('Invalid product: ' + id);
```
muffinresearch: Nice!

public/js/components/subscription.jsx
```diff
@@ -0,0 +1,42 @@
...
+        <div className="subnav">
```
muffinresearch: You probably want `<nav>` for this?

public/js/components/subscription.jsx
```diff
@@ -0,0 +1,42 @@
...
+          <div className="change-pay-account">
```
muffinresearch: Perhaps kill the extra divs?

public/scss/_subscriptions.scss
muffinresearch: Maybe put the clear on .subnav? You deffo don't want a div just for clearing.

muffinresearch:
r+wc

public/js/apps/management/app.jsx
kumar303: yeah, calling it with a list would be cool

public/js/actions/subscriptions.js
kumar303: this is an interesting case. It's actually not possible to put the spinner on the component alone because the data can be reloaded. When it gets reloaded, the component already exists (it's showing the old subscription data). I've seen the pattern of dispatching loading actions in other redux examples, I think it works. It's nice because the data reducer can reset itself when it gets a loading action.

public/js/reducers/user.js
kumar303: oh yeah, I'll update

public/js/components/subscription-list.jsx
kumar303: so it is! Fixed. Ugh, so many warnings.

public/js/components/subscription.jsx
kumar303: you so fancy

public/js/components/subscription.jsx
kumar303: the pain! make it stop! :fire: I tried styling the a tags as blocks and just couldn't get it to work.

kumar303:
> This PR means you win the job of moving this to an action

are you trying to clean up the world before we ship a working management interface?!

muffinresearch:
> are you trying to clean up the world before we ship a working management interface?!

No, only necessary things! Sadly moving the sub creation is needed for the form to be used for adding subs and adding payment methods independent of subs as discussed on vidyo.

PR #247, opened 2015-07-30
Add pay method linkage (fixes #222)

### LINKED ISSUE #222: Hookup add-card link to manage payment methods modal

muffinresearch:
* Renamed CardDetails -> AddSubscription
* Renamed GET_PAY_METHODS -> GOT_PAY_METHODS
* Add link to card form and show it. 
* Add addPayMethod action (endpoint needed for this to actually work)
* Add tests for pay-method actions.
* Add error handling for tokenization failures.
* Moved action tests into their own dir.


public/js/actions/pay-methods.js
```diff
@@ -0,0 +1,73 @@
...
+    jquery.ajax({
+      method: 'get',
+      url: '/api/braintree/mozilla/paymethod/',
+      context: this,
+    }).then(function(data) {
+      dispatch({
+        type: actionTypes.GOT_PAY_METHODS,
+        user: {
+          payMethods: data,
+        },
```
kumar303: When dispatching an action, I think it makes sense to keep it completely separate from the data layer. How about dispatching just `{type: ..., payMethods: data}` ? That way, the `user` part is not coupled to this at all.

public/js/actions/pay-methods.js
```diff
@@ -0,0 +1,73 @@
...
+          },
+          url: '/api/braintree/mozilla/paymethod/',
+          method: 'post',
+          dataType: 'json',
+        }).then(data => {
+          console.log('Successfully added a card');
+          dispatch({
+            type: actionTypes.GOT_PAY_METHODS,
+            management: {
+              tab: actionTypes.SHOW_ADD_PAY_METHOD,
```
kumar303: did you mean to dispatch this tab data? I don't see it used anywhere. It doesn't feel right to dispatch a tab change as part of this action. I'd think the management reducer could watch for GOT_PAY_METHODS and change the tab in response to it but I'd want to look at that in some detail.

public/js/actions/pay-methods.js
```diff
@@ -0,0 +1,73 @@
...
+            },
+            user: {
+              payMethods: data,
```
kumar303: again, `{payMethods: data}` without `user` ?

public/js/actions/subscriptions.js
```diff
@@ -38,18 +36,20 @@ export function getUserSubscriptions(jquery=$) {
 export function createSubscription(braintreeToken, productId,
                                    creditCard, jquery=$,
                                    BraintreeClient=braintree.api.Client) {
-  return (dispatch) => {
+  return dispatch => {
```
kumar303: :)

public/js/components/card-form.jsx
```diff
@@ -82,6 +82,7 @@ export default class CardForm extends Component {
   static defaultProps = {
     // This should always be overidden with a localized value.
     submitPrompt: 'Submit',
+    productId: PropTypes.string,
```
kumar303: why does CardForm need `productId` ? Maybe this is left over from fixing conflicts?

kumar303: `productId` isn't used

public/js/reducers/management.js
```diff
@@ -1,31 +1,47 @@
...
         debugMessage: action.error.debugMessage,
       },
     });
   }
 
-  if (action.type === actionTypes.GET_PAY_METHODS) {
-    return Object.assign({}, defaults, {
-      paymentMethods: action.management.paymentMethods,
+  if (action.type === actionTypes.SHOW_PAY_METHODS) {
+    console.log('management store: got action', action);
```
kumar303: to make these a bit more useful I've been adding a summary of the data reduction about to occur, like: 
````
console.log('management:

File: public/js/components/card-form.jsx (hunk 5 above)
Comment (muffinresearch) [inline public/js/components/card-form.jsx]: Hmm not sure I might have fumbled a conflict.

File: public/js/reducers/management.js [hunk 7]
```diff
@@ -1,31 +1,47 @@
 import * as actionTypes from 'constants/action-types';
 
 
-export default function management(state, action) {
-  console.log('management store: got action', action);
+export const initialMgmtState = {
+  error: null,
+  tab: null,
```
kumar303: do we not have a default tab? Oh, I guess this will change once we move to the non-accordion layout.

public/js/actions/pay-methods.js
muffinresearch: Yeah that might be possible. This doesn't work yet because no endpoint. So I'll tweak this as necessary once it's possible to make it work.

public/js/reducers/management.js
```diff
@@ -1,31 +1,47 @@
...
+
+  if (action.type === actionTypes.SHOW_ADD_PAY_METHOD) {
+    console.log('management store: got action', action);
+    return Object.assign({}, initialMgmtState, {
+      tab: action.type,
+    });
+  }
+
+  if (action.type === actionTypes.CREDIT_CARD_SUBMISSION_ERRORS) {
+    console.log('purchase: setting submission errors from', action);
```
kumar303: log message needs updating

public/js/reducers/management.js
```diff
@@ -1,31 +1,47 @@
...
+    return Object.assign({}, initialMgmtState, {
+      cardSubmissionErrors: action.apiErrorResult,
```
kumar303: hmm, I wonder if we should make this more generic like `submissionErrors`? I guess we'll figure it out later when we have other management tabs that need to accept error data.

public/js/reducers/management.js
```diff
@@ -1,31 +1,47 @@
...
     });
   }
 
   if (action.type === actionTypes.CLOSE_MODAL) {
-    return defaults;
+    console.log('management store: got action', action);
+    return initialMgmtState;
```
kumar303: I can foresee a situation where we want to close the modal but preserve the current tab. We can figure that out later though.

public/js/actions/pay-methods.js
muffinresearch: I'll remove the tab data for now.

public/js/views/add-pay-method.jsx
```diff
@@ -1,22 +1,46 @@
...
-  };
+    productId: PropTypes.string.isRequired,
+  }
+
+  componentDidMount() {
+    tracking.setPage('/add-pay-method');
+  }
+
+  handleCardSubmit(creditCard) {
+    console.log('submitting credit card to sign up for subscription',
```
kumar303: log message needs updating

public/js/views/add-subscription.jsx
```diff
@@ -6,7 +6,7 @@ import { default as tracking } from 'tracking';
 import { gettext } from 'utils';
 
 
-export default class CardDetails extends Component {
+export default class AddSubscription extends Component {
```
kumar303: :+1: :zap:

public/js/reducers/management.js
muffinresearch: Yeah, I've avoided things with modals since most of them are going to be removed.

tests/actions/test.subs-actions.js
```diff
@@ -107,6 +114,14 @@ describe('subscription actions', function() {
       });
     });
 
+    it('should dispatch an error on tokenization failure', function() {
+      var jquery = helpers.fakeJquery();
+      createSubscription(jquery.stub, FakeBraintreeClientError);
+      var action = dispatchSpy.firstCall.args[0];
+      assert.deepEqual(action,
+                       appActions.error('Braintree tokenization error'));
+    });
```
kumar303: The fact that there is a duplicate test for this makes me think we need a separate helper to wrap up the tokenization and common failure dispatch. It could be used maybe like this as a promise:
````
addSubscription(...) {
  return dispatch => {
    tokenizeCard(dispatch, braintreeToken, creditCard).then(nonce => {
      jquery.ajax(...)  // add the subscription
    });
  }
}
````

this way, we can make sure the helper always dispatches the tokenization failure.

If not in this patch, maybe file an issue for it?

File: tests/reducers/test.user-reducer.js [hunk 14]
```diff
@@ -108,3 +108,4 @@ describe('User Reducer', function() {
   });
 
 });
+
```
kumar303: file ends in new line. I forget why this is a problem though, hehe.

tests/actions/test.subs-actions.js
muffinresearch: Yeah I'd planned to dry that up in a future patch.

public/js/reducers/purchase.js
```diff
@@ -1,9 +1,9 @@
 import * as actionTypes from 'constants/action-types';
 
 
-export const initialState = {
+export const initialPurchaseState = {
   completed: false,
-  payment_methods: [],
+  payMethods: null,
```
kumar303: what's the point of null as opposed to an empty list? Just for consistency? It looks like it makes Purchase a little more tricky since the code has to defend against nulls. The reason null was used to initialize subscriptions was just to indicate that none had been loaded yet. For the purchase flow that wouldn't really apply (they are loaded at sign-in).

tests/actions/test.subs-actions.js
kumar303: filed: https://github.com/mozilla/payments-ui/issues/248

kumar303:
r+wc

public/js/reducers/purchase.js
muffinresearch: Because there's a difference between not having any payMethods e.g. `[]` and not having run the code to get them `null`. I'm not using that distinction yet but I was thinking it might be necessary if getPayMethods was needed in the future in addition to the way we get them at sign-in currently.

muffinresearch: Ah yes - I didn't realise this was purchase. null is unnecessary here.

PR #249, opened 2015-07-30
Rename 'purchase' -> 'transaction'

kumar303:


public/js/views/purchase.jsx
```diff
@@ -33,49 +33,48 @@ export default class Purchase extends Component {
...
   selectData(state) {
     return {
-      purchase: state.purchase,
+      transaction: state.transaction,
       user: state.user,
     };
   }
 
-  renderChild(result) {
+  renderChild(connector) {
```
muffinresearch: \o/

public/js/views/purchase.jsx
```diff
@@ -33,49 +33,48 @@ export default class Purchase extends Component {
...
-    } else if (result.purchase.payMethods &&
-               result.purchase.payMethods.length > 0) {
+    } else if (connector.transaction.availablePayMethods.length > 0) {
       console.log('rendering card listing');
       return (
         <CardListing
           productId={props.productId}
-          payMethods={result.user.payMethods}
-          {...bindActionCreators(purchaseActions, result.dispatch)}
+          payMethods={connector.transaction.availablePayMethods}
```
muffinresearch: `payMethods` is used in every other case.

muffinresearch:
I don't like `transactActions` because it looks too much like a typo - other than that r+wc

public/js/views/purchase.jsx
kumar303: Yeah, exactly. It suddenly dawned on me that we need to make a distinction here that they are only the *available* pay methods. I think this makes the logic read a lot better because nearby areas refer to `user.payMethods` for different purposes.

kumar303:
renamed to `transactionActions`

PR #251, opened 2015-07-30
Fix up API handling for adding a pay method

kumar303:
These changes are related to https://github.com/mozilla/payments-service/pull/121

This gets the action and data flow working after submitting a new card. 

However, the app gets stuck on the Add Card modal. I thought about this a little but I think it's complicated to fix in the current app. I say we hold off until the tab design layout.


muffinresearch:
r+

PR #252, opened 2015-07-31
Move subs with existing card to action (fixes #250)

### LINKED ISSUE #250: Submit handler in card choice needs to use the subscription action
See https://github.com/mozilla/payments-ui/blob/master/public/js/components/card-choice.jsx#L36

The form is setup to make a subscription based on the choice of an existing card.

muffinresearch:
Needed to do this on the way to deletion. 

yak

### DIFF
--- a/public/js/actions/subscriptions.js
+++ b/public/js/actions/subscriptions.js
@@ -33,43 +33,68 @@ export function getUserSubscriptions(jquery=$) {
 }
 
 
-export function createSubscription(braintreeToken, productId,
-                                   creditCard, jquery=$,
+export function createSubscription(productId, payMethod,
+                                   braintreeToken, jquery=$,
                                    BraintreeClient=braintree.api.Client) {
   return dispatch => {
 
-    var client = new BraintreeClient({
-      clientToken: braintreeToken,
-    });
+    function requestSub(opts={}) {
+      var data = {
+        plan_id: productId,
+      };
+      data.pay_method_uri = opts.payMethodUri;
+      data.pay_method_nonce = opts.nonce;
 
-    client.tokenizeCard({
-      number: creditCard.number,
-      expirationDate: creditCard.expiration,
-      cvv: creditCard.cvv,
-    }, (err, nonce) => {
-      if (err) {
-        console.error('Braintree tokenization error:', err);
-        dispatch(appActions.error('Braintree tokenization error'));
-      } else {
-        jquery.ajax({
-          data: {
-            pay_method_nonce: nonce,
-            plan_id: productId,
-          },
-          url: '/api/braintree/subscriptions/',
-          method: 'post',
-          dataType: 'json',
-        }).then(() => {
-          console.log('Successfully subscribed + completed payment');
-          dispatch(transactionActions.complete());
-        }).fail($xhr => {
+      return jquery.ajax({
+        data: data,
+        url: '/api/braintree/subscriptions/',
+        method: 'post',
+        dataType: 'json',
+      }).then(() => {
+        console.log('Successfully subscribed + completed payment');
+        dispatch(transactionActions.complete());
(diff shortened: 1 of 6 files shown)


public/js/actions/subscriptions.js
```diff
@@ -33,43 +33,68 @@ export function getUserSubscriptions(jquery=$) {
 }
 
 
-export function createSubscription(braintreeToken, productId,
-                                   creditCard, jquery=$,
+export function createSubscription(productId, payMethod,
+                                   braintreeToken, jquery=$,
                                    BraintreeClient=braintree.api.Client) {
```
kumar303: I think this is going to make the calling code too confusing. I suggest we move it to parameters. Luckily, we can use defaults, like this:
````
createSubscription({productId=undefined, payMethod=undefined, creditCard=undefined, 
                    jquery=$, BraintreeClient=braintree.api.Client}) {
}
````

this way you don't have to overload and introspect `payMethod`. It would be more explicit.

Comment (kumar303) [inline public/js/actions/subscriptions.js]: or to keep the other required args in place, it could be:
````
createSubscription(productId, braintreeToken, 
                   {payMethod=undefined, creditCard=undefined, 
                    jquery=$, BraintreeClient=braintree.api.Client}) {
}
````

Comment (kumar303) [inline public/js/actions/subscriptions.js]: note that without `({...} = {})` it makes the object a required parameter.

File: public/js/actions/subscriptions.js [hunk 2]
```diff
@@ -33,43 +33,68 @@ export function getUserSubscriptions(jquery=$) {
...
   return dispatch => {
 
-    var client = new BraintreeClient({
-      clientToken: braintreeToken,
-    });
+    function requestSub(opts={}) {
```
kumar303: It looks like this isn't optional so maybe just `requestSub(payMethodUri)` ? Or if you want to pass it as a parameter for explicitness it could be `requestSub({payMethodUri})` I think.

kumar303: it looks like nonce needs to be in there too: `requestSub({payMethodUri, nonce})`

kumar303: nit: while it's technically a request, `createSub()` would read better to me

public/js/actions/subscriptions.js
```diff
@@ -33,43 +33,68 @@ export function getUserSubscriptions(jquery=$) {
...
+      }).fail($xhr => {
+        if (data.pay_method_nonce) {
           dispatch({
             type: actionTypes.CREDIT_CARD_SUBMISSION_ERRORS,
             apiErrorResult: $xhr.responseJSON,
           });
-        });
+        } else {
+          dispatch(appActions.error('Subscription creation failed'));
+        }
```
kumar303: I don't think this if/else is in the right place. I think this should always dispatch CREDIT_CARD_SUBMISSION_ERRORS and the *receiver* of that data should decide how to handle it.

public/js/actions/subscriptions.js
```diff
@@ -33,43 +33,68 @@ export function getUserSubscriptions(jquery=$) {
...
+      });
+    }
+
+    // Check to see if payMethod is a card object
+    if (typeof payMethod === 'object') {
```
kumar303: I think using explicit params is better:
````
if (creditCard === undefined) {
  ...
} else if (payMethodUri === undefined) {
  ...
}
````

public/js/actions/subscriptions.js
kumar303: nevermind, I see that either `payMethodUri` or `nonce` can be used. I still think params would be nice: `createSub({payMethodUri=undefined, nonce=undefined})`

kumar303:
All good! Merge it in and I'll do some refactoring in https://github.com/mozilla/payments-ui/issues/254

PR #255, opened 2015-07-31
Add delete ui

muffinresearch:
This is a bit rough and ready and lacking in any new tests but it's the basic UI for delete card. 

I've moved the handlers up a level - I think there's more work to be done here to eliminate some of the overly nested components - but I think doing that will be best post ui-refactor.

### DIFF
--- a/public/js/actions/pay-methods.js
+++ b/public/js/actions/pay-methods.js
@@ -5,6 +5,33 @@ import * as actionTypes from 'constants/action-types';
 import * as appActions from './app';
 
 
+export function delPayMethod(payMethodUri, jquery=$) {
+
+  if (typeof payMethodUri === 'undefined') {
+    throw new Error('payMethodUri is undefined');
+  }
+
+  return dispatch => {
+    jquery.ajax({
+      method: 'post',
+      url: '/api/braintree/paymethod/delete/',
+      data: {
+        pay_method_uri: payMethodUri,
+      },
+      context: this,
+    }).then(function(data) {
+      dispatch({
+        type: actionTypes.GOT_PAY_METHODS,
+        payMethods: data.payment_methods,
+      });
+    }).fail(function() {
+      console.log('Deleting pay method failed');
+      dispatch(appActions.error('Deleting pay method failed'));
+    });
+  };
+}
+
+
 export function getPayMethods(jquery=$) {
   // Note: not currently used as payMethods
   // are added to user state upon login.
@@ -64,3 +91,4 @@ export function addCreditCard(braintreeToken, creditCard, jquery=$,
     });
   };
 }
+
--- a/public/js/apps/management/app.jsx
+++ b/public/js/apps/management/app.jsx
@@ -91,7 +91,6 @@ export default class ManagementApp extends Component {
           payMethods={connector.user.payMethods}
           {...boundMgmtActions}
           {...boundPayMethodActions}
-          {...boundUserActions}
         />
       ));
     }
--- a/public/js/apps/management/app.jsx
+++ b/public/js/apps/management/app.jsx
@@ -14,6 +14,7 @@ import { parseQuery } from 'utils';
 import Modal from 'components/modal';
 
 import AddPayMethod from 'views/add-pay-method';
+import DelPayMethod from 'views/del-pay-method';
 import BraintreeToken from 'views/braintree-token';
 import ModalError from 'views/modal-error';
 
@@ -84,6 +85,15 @@ export default class ManagementApp extends Component {
           />
         ));
       }
+    } else if (connector.management.tab === 'SHOW_DEL_PAY_METHOD') {
+      children.push((
+        <DelPayMethod
+          payMethods={connector.user.payMethods}
+          {...boundMgmtActions}
+          {...boundPayMethodActions}
+          {...boundUserActions}
+        />
+      ));
     }
 
     children.push(<Management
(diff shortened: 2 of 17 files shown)


public/js/views/del-pay-method.jsx
```diff
@@ -0,0 +1,49 @@
...
+  render() {
+    return (
+      <Modal handleClose={this.props.closeModal}
+        title={gettext('Delete Card')}>
+        <div>
+          <CardChoice
+            modifier='delete'
+            submitButtonText={gettext('Delete')}
+            submitButtonModifier='warning'
+            submitHandler={this.handleSubmit}
```
kumar303: huh, I'm surprised that `this` is the right context when this gets called.

public/js/reducers/management.js
```diff
@@ -31,6 +31,13 @@ export default function management(state, action) {
     });
   }
 
+  if (action.type === actionTypes.SHOW_DEL_PAY_METHOD) {
+    console.log('management store: got action', action);
```
kumar303: we can stop saying 'store' I guess

public/js/apps/management/app.jsx
```diff
@@ -84,6 +85,15 @@ export default class ManagementApp extends Component {
           />
         ));
       }
+    } else if (connector.management.tab === 'SHOW_DEL_PAY_METHOD') {
+      children.push((
+        <DelPayMethod
+          payMethods={connector.user.payMethods}
+          {...boundMgmtActions}
+          {...boundPayMethodActions}
+          {...boundUserActions}
```
kumar303: does it need user actions?

public/js/components/card-choice.jsx
```diff
@@ -33,7 +42,7 @@ export default class CardChoice extends Component {
   handleSubmit = (e) => {
     e.preventDefault();
     this.setState({isSubmitting: true});
-    this.props.createSubscription(this.props.productId, this.state.card);
+    this.props.submitHandler(e, this.state.card);
```
kumar303: why does the props function need to receive the event object? I don't see it being used in your implementation here.

public/js/components/submit-button.jsx
```diff
@@ -6,16 +6,21 @@ export default class SubmitButton extends Component {
 
   static propTypes = {
     isDisabled: PropTypes.bool,
+    modifier: PropTypes.string.isRequired,
```
kumar303: since there is a default value, this doesn't need to be required, right?

public/js/components/submit-button.jsx
```diff
@@ -6,16 +6,21 @@ export default class SubmitButton extends Component {
...
+    modifier: null,
+  }
+
   render() {
     var { isDisabled, text, showSpinner, ...buttonAttrs } = this.props;
 
     var buttonClassNames = cx({
       'spinner': showSpinner,
-    });
+    }, this.props.modifier);
```
kumar303: what does this do? Can it be called something more specific than `modifier` ?

public/js/views/del-pay-method.jsx
```diff
@@ -0,0 +1,49 @@
...
+  static propTypes = {
+    closeModal: PropTypes.func.isRequired,
+    delCreditCard: PropTypes.func.isRequired,
+    payMethods: PropTypes.arrayOf(
+      PropTypes.shape({
+        id: PropTypes.number,
+        resource_uri: PropTypes.string,
+        truncated_id: PropTypes.string,
+        type_name: PropTypes.string,
+      })
```
kumar303: since this is just passed down to CardChoice I don't think it makes sense to declare a full shape for it -- none of its properties are used here.

public/js/views/del-pay-method.jsx
kumar303: oh riiight, the magic arrow function

kumar303:
I ran out of time before I could work on the UI :) The delete pay methods patch turned out pretty epic :fried_shrimp: https://github.com/mozilla/payments-service/pull/123

I think this PR is missing the action to do deletion, right? I don't see it here. One thing that is different from solitude is I made the post param `pay_method_uri` to be more explicit. Docs are up: http://payments-service.readthedocs.org/en/latest/topics/api.html#post--api-braintree-paymethod-delete-

public/js/components/card-choice.jsx
muffinresearch: We're not using it currently so I've ripped that out.

public/js/components/submit-button.jsx
muffinresearch: It doesn't but then it's still required. If defaults were changed this would still stand.

public/js/components/submit-button.jsx
muffinresearch: Changed to cssModifier (modifier is a commonly used term for css classes added to modify the look of something).

public/js/views/del-pay-method.jsx
muffinresearch: Yeah, I was thinking that too - it gets passed down through several components. Seems a lot of repetition for nothing.

muffinresearch:
Going to land this to faciliate moving on with the UI changes. Added delPayMethod + some basic tests. It'll get cleaned up further once we're in the new world order.

public/js/actions/pay-methods.js
```diff
@@ -5,6 +5,33 @@ import * as actionTypes from 'constants/action-types';
 import * as appActions from './app';
 
 
+export function delPayMethod(payMethodUri, jquery=$) {
+
+  if (typeof payMethodUri === 'undefined') {
+    throw new Error('payMethodUri is undefined');
```
kumar303: this will cause a 400 in the API response which might be better because in some theoretical future we'll have a user friendly catch all for 400 responses. But this works for now, just probably worth making future actions like this "dumb" for simplicity.

PR #266, opened 2015-08-04
Work with a cross origin API so that webpack works

kumar303:
Supports https://github.com/mozilla/payments-service/issues/115

Other than supporting CORS (which required API fetching rework), I tied up a couple other loose ends:
* sign-in happens when the page loads now
* you see a spinner while the sign-in is getting processed
* I re-jigged the braintree token action/reducer
* I tried out a switch statement for the data reducers. Let me know how that looks/reads. I held off on converting others ;)
* I removed the Charles config but maybe you still want to stash it in the repo for safe keeping? Let me know.

Note that the CORS stuff requires the breaking API changes from https://github.com/mozilla/payments-service/pull/125 so I'll land these at the same time.


public/management.html
```diff
@@ -11,8 +11,8 @@
     <div class="spinner-cont">
       <div class="spinner"/></div>
     </div>
+    <script src="/bower_components/fxa-relier-client/fxa-relier-client.min.js"></script>
```
kumar303: This defines `FxaClient` so really this should happen before anything else.

tests/test.management.jsx
```diff
@@ -61,12 +68,4 @@ describe('Management', function() {
...
   });
 
-  it.skip('should show subscriptions on click', function() {
-    var view = mountView();
-    var button = helpers.findByClass(view, 'subs');
-
-    TestUtils.Simulate.click(button);
-    assert.equal(fakeSubscriptionGetter.called, true);
-  });
-
```
kumar303: I really can't see us ever needing this

public/js/actions/api.js
```diff
@@ -0,0 +1,30 @@
...
+import * as defaultSettings from 'settings';
+
+var csrfToken;
+
+
+export function fetch(request, {jquery=$, settings=defaultSettings} = {}) {
+  request = Object.assign({
+    dataType: 'json',
+    headers: {},
+    xhrFields: {withCredentials: true},
```
muffinresearch: Should this be conditional in some way?

public/js/actions/api.js
```diff
@@ -0,0 +1,30 @@
+import $ from 'jquery';
+
+import * as defaultSettings from 'settings';
+
+var csrfToken;
+
+
+export function fetch(request, {jquery=$, settings=defaultSettings} = {}) {
```
muffinresearch: Great idea!

public/js/actions/api.js
```diff
@@ -0,0 +1,30 @@
...
+  request.url = settings.apiPrefix + request.url;
+
+  if (csrfToken) {
+    request.headers['X-CSRFToken'] = csrfToken;
+  } else {
+    console.warn('making request to ' + request.url +
+                 ' but no CSRF token has been set');
+  }
+
+  return jquery.ajax(request);
```
muffinresearch: I wonder if we should have returned a standard promise here - would make switching jquery out even easier.

public/js/apps/management/main.js
```diff
@@ -1,9 +1,3 @@
-import $ from 'jquery';
 import { init as initApp } from './app';
 
-// Common ajax settings.
-$.ajaxSetup({
-  dataType: 'json',
-});
-
 initApp();
```
muffinresearch: If this is all this is doing I wonder if we should just kill the main.js files and make app.jsx the entry point for webpack.

public/js/reducers/management.js
```diff
@@ -4,78 +4,79 @@ import * as actionTypes from 'constants/action-types';
...
+export default function management(state, action) {
+  state = state || initialMgmtState;
+  var newState;
 
-  if (action.type === actionTypes.CREDIT_CARD_SUBMISSION_ERRORS) {
-    console.log('managment: setting submission errors from', action);
-    return Object.assign({}, initialMgmtState, {
-      cardSubmissionErrors: action.apiErrorResult,
-    });
+  switch (action.type) {
```
muffinresearch: This is much nicer.

public/js/settings.js
```diff
@@ -47,10 +47,27 @@ module.exports = {
...
+  apiPrefix: 'http://pay.dev:8000/api',
 };
+
+
+if (typeof window !== 'undefined') {
+  var host = window.location.host;
+  module.exports.fxaClientId = module.exports.fxaClientIdMap[host];
+
+  if (!module.exports.fxaClientId) {
+    console.warn('no FxA client ID has been configured for host ' + host);
```
muffinresearch: console.error?

tests/test.management.jsx
muffinresearch: :+1:

muffinresearch:
Really nice changes in here.

r+wc

public/js/actions/api.js
kumar303: maybe I should read the docs to see what it actually does :/

kumar303: The setting sends cookies on Access Control requests. I don't know what that fully means but it seems like maybe we shouldn't do that all the time. Good call. I changed it so it only does it when the app is running on a known host. I added some comments about it too.

public/js/actions/api.js
kumar303: yeah, I was thinking that too. We could use this function to get all the code ready for fetch. It will probably need to call fetch differently too (I think you pass in a URL). It could be done all in #84

public/js/settings.js
kumar303: I don't think so because there are some legit cases. One is that when we run the unit tests it will be served from localhost:XXXX which is fine. You see a warning at the start of the tests but no biggie.

public/js/apps/management/main.js
kumar303: can you make webpack call `initApp` directly? Less files is always good.

Issue #257, opened 2015-08-04
Use a logger middleware


gaearon:
Until you're ready with #244, you can remove console logs in reducers and instead add middleware like this:

```js
function logger({ getState }) {
  return next => action => {
    console.info('dispatching', action);
    const result = next(action);
    console.log('state after', getState());
    return result;
  };
}
```

kumar303:
oh nice! Thanks

PR #269, opened 2015-08-05
Use middleware for redux logging (#257)

kumar303:
for https://github.com/mozilla/payments-ui/issues/257

### DIFF
--- a/public/js/reducers/management.js
+++ b/public/js/reducers/management.js
@@ -10,73 +10,59 @@ export const initialMgmtState = {
 
 export default function management(state, action) {
   state = state || initialMgmtState;
-  var newState;
 
   switch (action.type) {
     case actionTypes.USER_SIGNED_IN:
       if (!state.tab) {
-        newState = Object.assign({}, initialMgmtState, {
+        return Object.assign({}, initialMgmtState, {
           tab: 'profile',
           view: actionTypes.SHOW_MY_ACCOUNT,
         });
       }
-      break;
+      return state;
     case actionTypes.APP_ERROR:
-      newState = Object.assign({}, initialMgmtState, {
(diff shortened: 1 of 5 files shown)


public/js/reducers/management.js
```diff
@@ -10,73 +10,59 @@ export const initialMgmtState = {
...
   switch (action.type) {
     case actionTypes.USER_SIGNED_IN:
       if (!state.tab) {
-        newState = Object.assign({}, initialMgmtState, {
+        return Object.assign({}, initialMgmtState, {
           tab: 'profile',
           view: actionTypes.SHOW_MY_ACCOUNT,
         });
       }
       break;
```
muffinresearch: Missed one.

public/js/reducers/management.js
```diff
@@ -10,73 +10,59 @@ export const initialMgmtState = {
...
-      break;
-  }
-
-  if (newState) {
-    console.log('management: new state for action', action);
+      return initialMgmtState;
   }
 
-  return newState || state;
+  return state;
```
muffinresearch: Could use a default case?

muffinresearch:
r+wc

public/js/reducers/management.js
kumar303: this technically is a reachable break because `state.tab` could be set. I thought it was a little nitpicky to remove all the unreachable breaks but eslint wouldn't pass without it :/

public/js/reducers/management.js
kumar303: ah yeah that might keep things more consistent

public/js/reducers/management.js
muffinresearch: Oh, yeah - makes sense. I missed the condition.

PR #270, opened 2015-08-05
Disable delete link when no cards (fixes #262)

### LINKED ISSUE #262: Disable delete button if there's no cards.
You can't delete something you don't have. We should just disable getting to that view.

muffinresearch:
menubar

### DIFF
--- a/public/js/views/management/pay-methods.jsx
+++ b/public/js/views/management/pay-methods.jsx
@@ -1,8 +1,9 @@
 import React, { Component, PropTypes } from 'react';
+import cx from 'classnames';
 
 import CardList from 'components/card-list';
 
-import { gettext } from 'utils';
+import { gettext, isDisabled } from 'utils';
 
 
 export default class PayMethods extends Component {
@@ -20,6 +21,10 @@ export default class PayMethods extends Component {
 
   handleDelPayMethod = e => {
     e.preventDefault();
+    if (isDisabled(e.target)) {
+      console.log('Delete link is disabled. no-op');
+      return;
+    }
     this.props.showDelPayMethod();
   }
 
@@ -35,6 +40,12 @@ export default class PayMethods extends Component {
   }
 
   render() {
+
+    var isDeleteDisabled = !this.props.payMethods ||
+      !this.props.payMethods.length;
+    var deleteClasses = cx('button', 'quiet', 'delete',
+                           {'disabled': isDeleteDisabled});
+
     return (
       <div>
         <h1>{gettext('Payment Methods')}</h1>
@@ -42,7 +53,7 @@ export default class PayMethods extends Component {
           {this.renderChild()}
           <a className="button quiet add-pay-method" href="#"
             onClick={this.handleAddPayMethod}>{gettext('Add a new card')}</a>
-          <a className="button quiet delete" href="#"
+          <a className={deleteClasses} href="#"
             onClick={this.handleDelPayMethod}>{gettext('Delete')}</a>
         </div>
       </div>
--- a/tests/test.utils.jsx
+++ b/tests/test.utils.jsx
@@ -27,34 +27,47 @@ describe('utils.parseQuery', function() {
 describe('utils.isDisabled', function() {
 
   it('should be disabled if disabled attr present', function() {
-    var lnk = document.createElement('a');
+    var button = document.createElement('button');
     var disabledAttr = document.createAttribute('disabled');
-    lnk.setAttributeNode(disabledAttr);
-    assert.equal(utils.isDisabled(lnk), true);
+    button.setAttributeNode(disabledAttr);
+    assert.equal(utils.isDisabled(button), true);
   });
 
   it('should be disabled if disabled="1"', function() {
-    var lnk = document.createElement('a');
-    lnk.setAttribute('disabled', 1);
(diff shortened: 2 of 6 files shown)


public/js/views/management/pay-methods.jsx
```diff
@@ -20,6 +21,10 @@ export default class PayMethods extends Component {
 
   handleDelPayMethod = e => {
     e.preventDefault();
+    if (isDisabled(e.target)) {
+      console.log('Link is disabled. no-op');
```
kumar303: could this be more specific like "not deleting pay methods because clicked button was disabled" ?

public/js/views/management/pay-methods.jsx
```diff
@@ -35,6 +40,12 @@ export default class PayMethods extends Component {
   }
 
   render() {
+
+    var isDeleteDisabled = !this.props.payMethods ||
```
kumar303: is it ok that this line doesn't end in a semi colon? Do we not need to put parentheses around the expression? Maybe babel fixes it all? ;)

public/js/views/management/pay-methods.jsx
```diff
@@ -35,6 +40,12 @@ export default class PayMethods extends Component {
...
+      !this.props.payMethods.length;
+    var deleteClasses = cx('button', 'quiet', 'delete',
+                           {'disabled': isDeleteDisabled});
```
kumar303: I think this would be less confusing as `isDisabled` but I see what you were doing.

public/js/views/management/pay-methods.jsx
```diff
@@ -42,7 +53,7 @@ export default class PayMethods extends Component {
           {this.renderChild()}
           <a className="button quiet add-pay-method" href="#"
             onClick={this.handleAddPayMethod}>{gettext('Add a new card')}</a>
-          <a className="button quiet delete" href="#"
+          <a className={deleteClasses} href="#"
```
kumar303: shouldn't you also do `<a ... disabled={isDisabled}>` here?

tests/test.utils.jsx
```diff
@@ -22,3 +22,39 @@ describe('utils.parseQuery', function() {
   });
 
 });
+
+
+describe('utils.isDisabled', function() {
+
+  it('should be disabled if disabled attr present', function() {
+    var lnk = document.createElement('a');
+    var disabledAttr = document.createAttribute('disabled');
```
kumar303: since this is a jsx file I'd assume you could also do `var lnk = <a disabled />;`

public/js/views/management/pay-methods.jsx
```diff
@@ -20,6 +21,10 @@ export default class PayMethods extends Component {
 
   handleDelPayMethod = e => {
     e.preventDefault();
+    if (isDisabled(e.target)) {
```
kumar303: this seems unfortunate because I feel like one day someone is going to forget to check for this. Is there really no react way to delete an element so that its event handlers aren't fired?

kumar303:
r+wc

public/js/views/management/pay-methods.jsx
muffinresearch: Sadly disabled attrs don't apply to links: http://www.w3.org/html/wg/drafts/html/master/semantics.html#concept-element-disabled

public/js/views/management/pay-methods.jsx
muffinresearch: Yeah I know what you mean. We'll have to be thorough! We need a click middleware.

I'd considered using pointer-events in CSS - I could add that too (as a saftety net) . But support isn't good enough for our needs http://caniuse.com/#feat=pointer-events - hence I thought we should just deal with it in the click. The other consideration was if we use pointer-events that would mask the problem in IE 9/10 because it would be fixed in everything else just not there.

Another possibility would be to replace the link with a button here since we're doing click events anyway - though that feels a little wrong.

I'll go ahead with this for now but let's add this to the to-discuss list (maybe before or after the standup).

public/js/views/management/pay-methods.jsx
muffinresearch: Afaik this is ok because this is an unambiguosly incomplete statement.

E.g: http://jsbin.com/hiqizu/edit?js,console 

Interestingly babel does also flatten this out looking at the demo output https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=foo%20%3D%20false%20%7C%7C%0A%20%20%20%20%20%20true%3B%0A

public/js/views/management/pay-methods.jsx
muffinresearch: I see I confused that in my tests. I'll fix them up to use a relevant element in the attr tests. I'll also make the class test only for links.

public/js/views/management/pay-methods.jsx
kumar303: yeah, I think it's probably safer for other browsers to flatten it. Cool, if babel does it I'm not worried :)

public/js/views/management/pay-methods.jsx
kumar303: sounds good. We'll just have to watch out for it in code reviews which is no biggie.

PR #272, opened 2015-08-05
Return to pay-methods after deletion/add-card (fixes #271)

### LINKED ISSUE #271: Fix navigation post adding/deleting a card
We need to navigate back to pay methods upon successful adding/removing a card.

muffinresearch:


kumar303:
r+wc

I think we should try this approach so I'm saying to land it but it tripped me up because I'm not sure it's the right thing to do.

The problem: we want to delete the payment methods *then* redirect to a list of payment methods. It's really two different actions but our caller (`DelPayMethod`) has no hook for when the action is complete so it cannot fire off more actions.

For the sake of argument, let's say we wanted to add an option for deleting a payment method in the purchase flow. The user could delete the payment method then return to the card choice listing to continue with the transaction. We couldn't reuse the `delPayMethod()` action in that case. We'd have to create a new action like `delPatMethodThenReturnToCardChoice()`.

One possible solution might be to return control to the calling component so that `DelPayMethod` could handle the delete button click more like:

````
  handleSubmit = (card) => {
    this.props.delPayMethod(card).then(() => this.props.showPayMethods());
  }
````

The bummer about that is we can no longer test `DelPayMethod` by doing a simple assert that `fakeDelPayMethod` was called; we'd have to make the test async.

Thoughts?

If we keep the async logic *only* in the actions themselves then we only have to do an async test for the actions. If we return control to the calling component (my example above) then we'd have to do async testing in two places. This makes me lean more towards your approach of doing the double dispatch and just know that we'd have to create more specific actions in the future instead of re-using actions as is.

Maybe the `delPayMethod` action should at least move to `actions/management` ?

kumar303:
Also, I searched for a little advice on this and it seems like this thread is saying it's ok to fire off multiple dispatches from an action https://github.com/gaearon/redux/issues/199

muffinresearch:
Yeah I had similar concerns. I started trying to push the handling to the reducer but that felt really wrong. 

This seemed simple and just works (now) but I agree it's hard-wiring the navigation into that action. But it gets the job done in lieu of a more robust solution (aka a quick fix) :).

We should have a think in general about navigation patterns for redux and how do deal with it more generally. This all feels a bit too low level. Maybe the original action caller (e.g. what's asking for the delete) should be able to inform what happens on success - that would give us more flexibility and fairly easily? I have also been thinking it would be nice to introduce urls to the management interface before we get too far in.

I'll land this for now but let's have a chat about it. Fwiw I recently joined #redux in the reactiflux slack group and that might be a good place to asking questions about what other people are doing to solve these kind of problems.

PR #274, opened 2015-08-06
Move subscription className (fixes #263)

### LINKED ISSUE #263: Look to re-instate class on subscriptions.
#258 moved the subscription className to the li. As the subscription component might be re-used independently this should be re-instated.

Anything using the li.subscription could just use the element selector instead.

muffinresearch:


kumar303:
r+

thanks

PR #276, opened 2015-08-06
Remove conditional around fetching subs (fixes #275)

### LINKED ISSUE #275: Subs should always be fetched
Otherwise we don't get new data.

In the future we should have new subs pushed. :)

muffinresearch:
This means we always fetch subs when clicking the nav for subscriptions.


kumar303:
r+wc

Looks like the right change to make but while you're in there, might as well add a quick test that mounts the view and asserts the getter property is called. It will be helpful at the very least just to get coverage of mounting this component.

PR #281, opened 2015-08-06
Add drop-down via React (fixes #267)

### LINKED ISSUE #267: Implement card drop-down in react
The prototype is here: 

http://output.jsbin.com/suyama/latest

Next step is to implement the same widget in react.

muffinresearch:
This is the drop-down component in React that follows the mocked-up version.
It manages it's own internal state. Values are passed up to parents via passing in a onCardChange prop. This way the parent can get notified of changes if it wants to and can also maintain state this way like it does for the card choice.

Dealing with a default selection will be a todo. Can work on that when we have the card default api sorted. We'll probably want to have the 'Please select' default option removed for that if we're going to keep it for this part of the UI.

Adding this to the Payment methods view is mostly a strawman so I can use it as a discussion point with Bram. Reversing this back should be easy though so it's not a big deal if we want to go back to a card list here.

http___pay_webpack_8080_webpack-dev-server_management_html_access_token_2434814ed1502d9d3d1d5d7eb3f0280f75b99f00094cc4645b00cb008a7a5f30_is_not_available

### DIFF
--- a/public/js/components/card-drop-down.jsx
+++ b/public/js/components/card-drop-down.jsx
@@ -1,0 +1,110 @@
+import React, { Component, PropTypes } from 'react';
+import cx from 'classnames';
+
+import CardIcon from 'components/card-icon';
+import { gettext } from 'utils';
+
+const defaultSelectText = gettext('Please select');
+
+
+export default class CardDropDown extends Component {
+
+  static propTypes = {
+    cards: PropTypes.arrayOf(
+      PropTypes.shape({
+        id: PropTypes.number,
+        resource_uri: PropTypes.string,
+        truncated_id: PropTypes.string,
+        type_name: PropTypes.string,
+      })
+    ).isRequired,
+    cssModifier: PropTypes.string,
+    onCardChange: PropTypes.func.isRequired,
+  }
+
+  constructor(props) {
+    super(props);
+    this.state = {
+      selectedText: defaultSelectText,
+      selectedCardType: null,
+      isFocused: false,
+    };
+  }
+
+  handleFocus = () => {
+    this.setState({isFocused: true});
+  }
+
+  handleBlur = () => {
+    this.setState({isFocused: false});
+  }
+
+  handleChange = e => {
+    if (this.props.onCardChange) {
+      this.props.onCardChange(e);
+    }
+    var selectedText;
+    var selectedNode = e.target.options[e.target.selectedIndex];
+    if (selectedNode) {
+
+      var selectedCardType = selectedNode.getAttribute('data-type');
+      if (selectedCardType) {
+        selectedText = selectedNode.firstChild.nodeValue.replace(
+          selectedCardType.toUpperCase(), '');
+      }
+    }
+    this.setState({
+      selectedText: selectedText || defaultSelectText,
+      selectedCardType: selectedCardType,
+    });
+  }
+
+  render() {
+    var cardData = this.props.cards;
+    var cardOptions = [];
+
+    cardOptions.push(<option value="">{defaultSelectText}</option>);
+
+    for (var i = 0; i < cardData.length; i += 1) {
+      var { selected, ...card } = cardData[i];
+      var optionText = card.type_name.toUpperCase() +
+        ' ●●●● ●●●● ●●●● ' + card.truncated_id;
+
+      cardOptions.push(
+        <option
+          data-type={card.type_name.toLowerCase()}
+          key={card.id}
+          selected={selected}
+          value={card.resource_uri}
+        >{optionText}</option>
+      );
+    }
+
+    var contentClasses = cx('content', {
+      'has-card': this.state.selectedCardType,
+    });
+
+    var proxySelectClasses = cx('proxy-select', {
+      'active': this.state.isFocused,
+    });
+
+    return (
+      <div className={proxySelectClasses}>
+        {this.state.selectedCardType ?
+         <CardIcon cardType={this.state.selectedCardType} /> :
+         null}
+        <span className={contentClasses} ariaHidden="true">
+          <span className="vh">{this.state.selectedCardType}</span>
+          {this.state.selectedText}
+        </span>
+        <select
+          onBlur={this.handleBlur}
+          onChange={this.handleChange}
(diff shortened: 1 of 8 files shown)


public/js/components/card-drop-down.jsx
```diff
@@ -0,0 +1,112 @@
...
+    this.setState({isFocused: false});
+  }
+
+  handleChange = e => {
+    if (this.props.onCardChange) {
+      this.props.onCardChange(e);
+    }
+    var selectedText;
+    var selectedNode = e.target.options[e.target.selectedIndex];
+    console.log(selectedNode);
```
kumar303: still needed? If so, it could do with a preamble

public/js/components/card-drop-down.jsx
```diff
@@ -0,0 +1,112 @@
...
+    this.setState({isFocused: true});
+  }
+
+  handleBlur = () => {
+    this.setState({isFocused: false});
+  }
+
+  handleChange = e => {
+    if (this.props.onCardChange) {
+      this.props.onCardChange(e);
```
kumar303: this seems a little limiting. Should it receive `selectedCardType` too maybe? We could always expand upon it later.

public/js/components/card-drop-down.jsx
```diff
@@ -0,0 +1,112 @@
...
+    }
+    var selectedText;
+    var selectedNode = e.target.options[e.target.selectedIndex];
+    console.log(selectedNode);
+    if (selectedNode) {
+
+      var selectedCardType = selectedNode.getAttribute('data-type');
+      console.log('selectedCardType: ' + selectedCardType);
+      if (selectedCardType) {
+        selectedText = selectedNode.firstChild.nodeValue.replace(
```
kumar303: fail-kids

kumar303: I'm pretty sure this is a strict no-no in react. I would assume instead you could just render the `{optionText}` differently once you receive `selectedCardType` in the state change, no? Did that not work?

kumar303: hmm, it looks like you're maybe already doing that. Was it not enough?

public/js/components/card-drop-down.jsx
```diff
@@ -0,0 +1,112 @@
...
+      selectedText: selectedText || defaultSelectText,
+      selectedCardType: selectedCardType,
+    });
+  }
+
+  render() {
+    var cardData = this.props.cards;
+    var cardOptions = [];
+
+    cardOptions.push(<option value="">{gettext('Please Select')}</option>);
```
kumar303: can you just use `{defaultSelectText}` here?

public/js/components/card-drop-down.jsx
kumar303: oh wait, this isn't modifying the DOM? It's just retrieving a value? Sorry! Ignore me.

public/js/components/card-drop-down.jsx
```diff
@@ -0,0 +1,112 @@
...
+    var proxySelectClasses = cx('proxy-select', {
+      'active': this.state.isFocused,
+    });
+
+    return (
+      <div className={proxySelectClasses}>
+        {this.state.selectedCardType ?
+         <CardIcon cardType={this.state.selectedCardType} /> :
+         null}
+        <span className={contentClasses} ariaHidden>
```
kumar303: :+1: accessible

kumar303: do you need to specify `ariaHidden=true` or `ariaHidden=false` ?

public/js/views/management/pay-methods.jsx
```diff
@@ -1,7 +1,8 @@
 import React, { Component, PropTypes } from 'react';
 import cx from 'classnames';
 
-import CardList from 'components/card-list';
+//import CardList from 'components/card-list';
```
kumar303: :fries:

tests/test.card-drop-down.jsx
```diff
@@ -0,0 +1,109 @@
...
+  it('updates internal state when card is selected', function() {
+    var event = {
+      target: {
+        value: '/braintree/mozilla/paymethod/2/',
+        selectedIndex: 2, // This is 2 because of the default option.
+        options: helpers.findAllNodesByTag(this.CardDropDown, 'option'),
+      },
+    };
+    this.CardDropDown.handleChange(event);
+    assert.ok(this.CardDropDown.setState.calledWith({
```
kumar303: do you need a spy? Couldn't you just check `this.CardDropDown.state` ?

kumar303: did you consider a test that looked for the selected option instead? Checking the state seems like an implementation detail.

public/js/components/card-drop-down.jsx
kumar303: nevermind, I guess you can get it from `e.target`

public/js/components/card-drop-down.jsx
muffinresearch: Ah yes, you're right this should be "true"

tests/test.pay-methods.jsx
```diff
@@ -28,13 +28,8 @@ describe('PayMethods', function() {
...
 
-  it('should have a visa card present', function() {
-    var content = helpers.findByClass(this.PayMethods, 'cctype-visa');
-    assert.ok(TestUtils.isDOMComponent(content));
-  });
-
-  it('should have an amex card present', function() {
-    var content = helpers.findByClass(this.PayMethods, 'cctype-amex');
+  it('should have a card dropdown', function() {
+    var content = helpers.findByClass(this.PayMethods, 'proxy-select');
```
kumar303: you could also check the component type, no? That might be better in case someone wants to change the css classes later.

kumar303:
I think I was confused about manipulating DOM, ignore that.

r+wc

tests/test.card-drop-down.jsx
muffinresearch: Yeah I think this was all the bits I wanted to test. It's a bit tricky dealing with looking in at the select because the events are faked.

PR #294, opened 2015-08-06
add in a title

### LINKED ISSUE #286: Need a decent title
The <title> for the management interface is just "Management", it should be something related to the function that the page is on.

andymckay:
* super simple setTitle as discussed, will probably be removed later when the data structure for pages lands
* remove prefix for now as per comments
* fixes #286

### DIFF
--- a/public/js/utils.jsx
+++ b/public/js/utils.jsx
@@ -41,6 +41,11 @@ export function gettext(string){
 }
 
 
+export function setTitle(string, doc=document) {
+  doc.title = string;
+}
+
+
 export function parseQuery(url){
   //
   // Given a complete URL, parse the query string and return an
--- a/public/js/views/management/history.jsx
+++ b/public/js/views/management/history.jsx
@@ -1,9 +1,13 @@
 import React, { Component } from 'react';
 
-import { gettext } from 'utils';
+import { gettext, setTitle } from 'utils';
 
 
 export default class History extends Component {
+
+  componentDidMount() {
+    setTitle(gettext('Receipts and history'));
+  }
 
   render() {
     return (
--- a/tests/test.utils.jsx
+++ b/tests/test.utils.jsx
@@ -71,3 +71,13 @@ describe('utils.isDisabled', function() {
   });
 
 });
+
+
+describe('utils.setTitle', function() {
+
+    it('should set the title', function() {
+        utils.setTitle('Pay', document);
+        assert.equal(document.title, 'Pay');
+    });
+
+});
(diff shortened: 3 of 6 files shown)


public/js/utils.jsx
```diff
@@ -41,6 +41,11 @@ export function gettext(string){
 }
 
 
+export function setTitle(string) {
+  document.title = gettext('Pay') + ' | ' + string;
```
muffinresearch: If we want to be consistent with FxA then we could totally drop the prefix.

public/js/views/management/history.jsx
```diff
@@ -5,6 +5,10 @@ import { gettext } from 'utils';
 
 export default class History extends Component {
 
+  componentDidMount() {
+    setTitle(gettitle('My account'));
```
muffinresearch: This should be "Receipts and history"

public/js/utils.jsx
```diff
@@ -41,6 +41,11 @@ export function gettext(string){
 }
 
 
+export function setTitle(string) {
```
muffinresearch: Perhaps add a basic test for this. Maybe inject the document object that way you can check the property is set. You can now use default args `setTitle(string, doc=document)`.

public/js/views/management/history.jsx
muffinresearch: Presumably `gettitle` shoudl be `gettext`.

muffinresearch: You're missing the import of `setTitle` here.

public/js/utils.jsx
andymckay: Marketplace does the pipe but does it as a suffix. Personally I'm not tied to it, blah lets remove it.

tests/test.utils.jsx
```diff
@@ -71,3 +71,13 @@ describe('utils.isDisabled', function() {
   });
 
 });
+
+
+describe('utils.setTitle', function() {
+
+    it('should set the title', function() {
+        utils.setTitle('Pay', document);
```
muffinresearch: This is where you pass in your own object so that you don't manipulate the real title.

muffinresearch:
r+wc

public/js/utils.jsx
```diff
@@ -41,6 +41,11 @@ export function gettext(string){
 }
 
 
+export function setTitle(string, doc=document) {
+  doc.title = string;
```
muffinresearch: I like this, I think it makes it simple e.g. no prefix/suffix or pipe. :+1:

PR #295, opened 2015-08-07
Add tabzilla (fixes #285)

### LINKED ISSUE #285: Missing mozilla branding
Most (if not all) Mozilla properties use tabzilla to have mozilla in the top right hand corner. The spiffy new management interface doesn't have anything for that.

It also doesn't have a footer that contains required links like: privacy policy and terms of use.

Should we have those sorts of features accessible somewhere.

muffinresearch:
Looks like this:

mozilla_firefox_and_terminal_ _ssh_ _80x40


tofumatt:
r+

PR #298, opened 2015-08-07
Re-org tests under dirs (fixes #297)

### LINKED ISSUE #297: Re-organize tests by type
We've started having the tests under different dirs to represent what they're testing we should continue this through to dealing with components and views too.

muffinresearch:


muffinresearch:
Landing this as it's a trivial change.

kumar303:
yep, looks good :)

PR #301, opened 2015-08-10
Allow default selection and default option to be off (fixes #302)

### LINKED ISSUE #302: Make card dropdown more flexible

muffinresearch:
* Allow the default "Please select" option to be turned off
* Allow for a default selection.
* Move the tests to the right dir.


public/js/components/card-drop-down.jsx
```diff
@@ -20,15 +20,51 @@ export default class CardDropDown extends Component {
     ).isRequired,
     cssModifier: PropTypes.string,
     onCardChange: PropTypes.func.isRequired,
+    selectedCardResource: PropTypes.string,
+    showDefaultOption: PropTypes.bool.isRequired,
```
kumar303: since this has a default it probably isn't required

public/js/components/card-drop-down.jsx
```diff
@@ -20,15 +20,51 @@ export default class CardDropDown extends Component {
...
+
+  static defaultProps = {
+    showDefaultOption: true,
   }
 
   constructor(props) {
     super(props);
-    this.state = {
+    var cardData;
+    // Get the data for a the default selected card.
```
kumar303: typo in this comment

public/js/components/card-drop-down.jsx
```diff
@@ -20,15 +20,51 @@ export default class CardDropDown extends Component {
...
+    if (props.selectedCardResource) {
+      cardData = this.getSelectedCard(props.selectedCardResource);
```
kumar303: this fetches data from the API, right? I don't know if the constructor is the right place since I think React creates components then caches them. Shouldn't this go in `componentWillMount()` instead?

public/js/components/card-drop-down.jsx
```diff
@@ -20,15 +20,51 @@ export default class CardDropDown extends Component {
...
+    }
+
+    // If there's default selected option and the defaultOption is off.
+    // we set the "fakeSelect" to the first option.
+    if (!cardData && !props.defaultOption) {
```
kumar303: I don't see `defaultOption` in prop types, only `showDefaultOption`

public/js/components/card-drop-down.jsx
```diff
@@ -20,15 +20,51 @@ export default class CardDropDown extends Component {
...
     super(props);
-    this.state = {
+    var cardData;
+    // Get the data for a the default selected card.
+    if (props.selectedCardResource) {
+      cardData = this.getSelectedCard(props.selectedCardResource);
+    }
+
+    // If there's default selected option and the defaultOption is off.
+    // we set the "fakeSelect" to the first option.
```
kumar303: is `fakeSelect` a thing? Or did you mean `"fake select"` :)

public/js/components/card-drop-down.jsx
```diff
@@ -20,15 +20,51 @@ export default class CardDropDown extends Component {
...
+    if (!cardData && !props.defaultOption) {
+      cardData = props.cards[0];
```
kumar303: should this guard against an empty `props.cards` list?

public/js/components/card-drop-down.jsx
```diff
@@ -20,15 +20,51 @@ export default class CardDropDown extends Component {
...
+    var initialState = {
       selectedText: defaultSelectText,
       selectedCardType: null,
       isFocused: false,
     };
+
+    if (cardData) {
+      initialState.selectedText = '●●●● ●●●● ●●●● ' + cardData.truncated_id;
+      initialState.selectedCardType = cardData.type_name.toLowerCase();
+    }
```
kumar303: I think it would be cleaner if a lot of this default selection code were to reuse `updateCardSelection()` even if it means pulling something out of there into a helper method.

public/js/components/card-drop-down.jsx
```diff
@@ -20,15 +20,51 @@ export default class CardDropDown extends Component {
...
+
+    this.state = initialState;
+  }
+
+  getSelectedCard(selectedCardResource) {
+    var selectedCard = this.props.cards.filter(function(item) {
+      return (item.resource_uri === selectedCardResource);
+    });
```
kumar303: could be `filter(item => item.resource_uri === selectedCardResource)`

public/js/components/card-drop-down.jsx
```diff
@@ -39,40 +75,46 @@ export default class CardDropDown extends Component {
...
   render() {
     var cardData = this.props.cards;
     var cardOptions = [];
 
-    cardOptions.push(<option value="">{defaultSelectText}</option>);
+    if (this.props.showDefaultOption === true) {
+      cardOptions.push((
+        <option
+          value=""
+          key="def"
```
kumar303: what does `def` mean? Can this key be made more unique?

public/js/components/card-drop-down.jsx
```diff
@@ -39,40 +75,46 @@ export default class CardDropDown extends Component {
...
+          ref="def">{defaultSelectText}</option>)
```
kumar303: nit: this trailing parenthesis looks off to me because it doesn't match the opening parentheses style. Only mentioning it because I stared at it a while thinking you'd forgotten a semi colon.

public/js/components/card-drop-down.jsx
kumar303: I guess this option would never be inserted twice? So maybe it's ok

kumar303: ah, def = default. Can it just be `key="default"` ?

kumar303: oh, heh, it's a reserved word, right. I would prefer `ref="_default"` even, just something readable.

kumar303:
I don't fully understand the fake select thing but, hey, whatever works.

r+wc

public/js/components/card-drop-down.jsx
muffinresearch: No this doesn't fetch any data.

public/js/components/card-drop-down.jsx
muffinresearch: Oh shiny.

public/js/components/card-drop-down.jsx
muffinresearch: It can't directly (I did that originally) because you can't call `setState` in a constructor. I can certainly DRY it up a bit though.

Issue #279, opened 2015-08-06
Initial Receipts and History (transactions) view in management interface


kumar303:
Start a grid view of transaction history fetched from the API

PR #303, opened 2015-08-10
Initial Receipts and History view (#279)

kumar303:
screenshot 2015-08-10 09 59 57
Supports: https://github.com/mozilla/payments-ui/issues/279


public/js/views/management/history.jsx
```diff
@@ -1,18 +1,90 @@
...
+        return <p>{gettext("You haven't purchased anything yet.")}</p>;
+      }
+    }
   }
 
   render() {
     return (
       <div>
-        <h1>{gettext('Receipts and history')}</h1>
+        <h1>{gettext('Transaction Receipts & History')}</h1>
```
muffinresearch: Everything tends to only upper case the first word of the title. So this should probably be:

`Transaction receipts & history`

Does 'and' not fit? I'll make the sidebar more flexible in due course but as a quick fix you could make it wider if you want to use `and` instead of `&`

public/js/views/management/history.jsx
```diff
@@ -1,18 +1,90 @@
...
 
 export default class History extends Component {
 
+  static propTypes = {
+    getUserTransactions: PropTypes.func.isRequired,
+    transactions: PropTypes.array.isRequired,
+  };
+
   componentDidMount() {
     setTitle(gettext('Receipts and history'));
```
muffinresearch: Probably want this to match the h1.

public/js/views/management/history.jsx
kumar303: The UX had it like this in the mock http://payments.readthedocs.org/en/latest/_images/ux-management-wireframe-wide.png Should we do it differently?

public/js/views/management/history.jsx
kumar303: ah right

public/js/views/management/history.jsx
muffinresearch: Ah ok - I'll and to the list to ask bram if there's a rule we should follow. I'd expect these to be consistent.

muffinresearch:
Lgtm r+wc

Issue #278, opened 2015-08-06
Get management sign-in working for AWS instance


kumar303:
this requires a patch to determine the FxA client ID by host

PR #317, opened 2015-08-11
Elaborate on per host settings to get AWS working (#278)

kumar303:
Supports https://github.com/mozilla/payments-ui/issues/278


andymckay:
r+

PR #322, opened 2015-08-12
Split css into common + app specific (fixes #172)

### LINKED ISSUE #172: Add separate css target for common, payment and management interface.
This is so we can share common css and keep app specific css just for each app.

Later we can consider moving to using the css features in webpack which can put the css into the js.

muffinresearch:


public/scss/_transaction.scss
```diff
@@ -0,0 +1,3 @@
+body {
+    padding-top: 30px;
```
muffinresearch: Fixes padding specific to transaction flow.

kumar303:
This seems fine for now but couldn't we just merge `common` into both bundles? That way we wouldn't have to link to `common.css`.

r+wc

kumar303:
I guess by making a separate http request to `common.css` we optimize a future visit to the management app by priming the cache. I dunno, I think we should prioritize optimization of the buy flow because that's the money path :) We should do anything and everything to speed up the buy flow.

muffinresearch:
> This seems fine for now but couldn't we just merge common into both bundles? That way we wouldn't have to link to common.css.

Yeah it's also about drying up the config as much as anything. We might even switch to module loading the css and then webpack will work it out for us.

PR #325, opened 2015-08-12
Rename card to pay method (fixes #310)

### LINKED ISSUE #310: s/Card/PayMethod/
The expectation is the Radio button choice or the DropDown will be used for payMethods including account pay methods e.g. paypal.

As a result of that it's going to be a good idea to move all things to refer to payMethods instead of cards.

CardChoice -> PayMethodChoice
CardListing -> PayMethodListing
CardIcon -> PayMethodIcon

etc.

muffinresearch:
Ignore the first commit it's a dep of this branch.

This branch renames card/Card  -> payMethod/PayMethod where the components views are actually all about payment methods not just cards.


public/js/components/pay-method-icon.jsx
```diff
@@ -0,0 +1,37 @@
+import cx from 'classnames';
+import React, { Component, PropTypes } from 'react';
+
+// Just a convenience mapping for payMethods from payMethod-validator
+// to shorted classes used in CSS.
+const payMethodTypeMap = {
+  'american-express': 'amex',
+  'diners-club': 'diners',
+  'master-card': 'mastercard',
+};
```
kumar303: hmm, this is going to need a lot of refactoring to wedge in paypal ;) I guess it's better to at least prepare by calling it `PayMethodIcon`.

muffinresearch: This specific is just a mapping for being able to handle cardTypes handed to us by the card validator which have slight variants on the names for cards we use internally. 

A paymethod that's an account will just be handled slightly differently it's still essentially all the same thing.

public/js/components/pay-method-choice.jsx
```diff
@@ -27,36 +27,36 @@ export default class CardChoice extends Component {
     super(props);
     this.state = {
       isSubmitting: false,
-      card: (this.props.cards.length === 1 ?
-             this.props.cards[0].resource_uri : null),
+      payMethod: (this.props.payMethods.length === 1 ?
+             this.props.payMethods[0].resource_uri : null),
```
kumar303: needs an indent

public/js/components/submit-button.jsx
```diff
@@ -7,10 +7,10 @@ import { gettext } from 'utils';
 export default class SubmitButton extends Component {
 
   static propTypes = {
-    cssModifier: PropTypes.string.isRequired,
+    cssModifier: PropTypes.string,
```
kumar303: I guess it's not fallout from this patch but I just noticed the `defaultProps` have the wrong property name for `cssModifier`

muffinresearch: Yep, just noticed that in the console.

kumar303:
Nice one. r+wc

Issue #314, opened 2015-08-11
Implement management sign out


kumar303:
We have [new mocks](http://payments.readthedocs.org/en/latest/_images/ux-management-wireframe-wide.png) for sign out so let's hook it up. There are some skipped tests for how it used to work (with the old button) so the code is pretty much all there. The sign out data state should already happen naturally. The view at sign out time will be the initial sign-in page as if you loaded the app for the first time.

muffinresearch:
Styles are landed so this is ready to go.

Issue #287, opened 2015-08-06
Remove access token from the URL


andymckay-limited-access:
Access tokens are now short lived, but I don't think we should keep the access token around in the URL, for the simple reason that any saving or sharing of the URL to the management interface will now have the access token in, something we should discourage.

This might be related to #277.

PR #328, opened 2015-08-12
Gets sign-out working in management (#314)

kumar303:
sign-out mov

supports https://github.com/mozilla/payments-ui/issues/314


tests/views/test.sign-in.jsx
```diff
@@ -0,0 +1,55 @@
...
+  });
+
+  it('should do a token sign-in with access token', function() {
+    var token = 'some-fxa-token';
+    mountView({accessToken: token});
+    assert.equal(fakeTokenSignIn.firstCall.args[0], token);
+  });
+
+  it('should do a user sign-in without access token', function() {
+    mountView({accessToken: null});
```
kumar303: You may know this but I just learned something cool here. If I were to do `mountView({accessToken: undefined})` then the value of `accessToken` is set to the default, `fakeAccessToken`. That's nice. It's not what I wanted for this case but I actually didn't know that was possible. You can't do that in python; passing `None` would always set it to `None`.

public/index.html
```diff
@@ -16,6 +16,7 @@
         </div>
       </div>
     </main>
+    <script src="/bower_components/fxa-relier-client/fxa-relier-client.min.js"></script>
```
kumar303: It seemed natural in this patch to allow a user sign in for the main buy flow if they happen to land on the URL without an access token (by editing the URL or whatever). The sign in part works but immediately after that the buy flow fails on the console because there isn't a product ID in the query string. We could improve that error later I suppose. I toyed with disallowing a sign-in on the buy flow but it introduced a lot of extra code.

andymckay:
Noted that this works if you sign in and sign out from the management interface.

If you sign in from payments-example then sign out from the management interface it doesn't work. 

Wondering if that's related to the access_token in the URL, which occurs when you sign in from payments-example, and I commented on in #287

andymckay:
Video of said issue: https://www.dropbox.com/s/bbl1du9axlqmipi/sign-out-management.mov?dl=0

andymckay:
The sign-in message on sign-out seems pretty consistent, would be cool to fix that up, could we file a seperate bug for that?

muffinresearch:
> During sign-out in this screencast it looks like it's saying "signing in." That's only because the sign-out happened really fast and it is already preparing the sign in form.

We should probably make that default to `loading` e.g. `<Spinner/>` then it won't matter.

tests/helpers.jsx
```diff
@@ -180,10 +180,9 @@ export function fakeFetch({xhrError={}, returnedData={},
 
 
 export function stubComponent() {
-  return React.createClass({
-    displayName: 'StubComponent',
-    render: function() {
-      return <div></div>;
-    },
-  });
+  return class StubComponent extends Component {
```
muffinresearch: \o/

tests/views/test.sign-in.jsx
muffinresearch: Yeah that makes sense because that's the same thing as a key not being present at all.

Seeing the exported transpilation is quite helpful for visualisation:

https://babeljs.io/repl/#?experimental=true&evaluate=true&loose=false&spec=false&code=function%20mountView%28{accessToken%3DfakeAccessToken%2C%20user%3DfakeUser}%20%3D%20{}%29%20{%0A%20%20%0A}

public/index.html
muffinresearch: I'm not sure about that. It's more code to load on the buy flow and direct linking is not expected afaik.
Can we not just hard fail early if conditions aren't met?

tests/apps/test.management-app.jsx
```diff
@@ -53,10 +56,13 @@ describe('management/app', function() {
       },
     });
 
-    var child = TestUtils.findRenderedComponentWithType(
+    var mg = TestUtils.findRenderedComponentWithType(
```
muffinresearch: 'mg' is a bit short.  'mgmt' is a bit more immediate.

public/js/views/management.jsx
```diff
@@ -70,14 +54,9 @@ export default class Management extends Component {
...
-  userSignIn = e => {
-    e.preventDefault();
-    this.props.userSignIn();
-  }
-
-  userSignOut = e => {
+  showSignOut = e => {
     e.preventDefault();
-    this.props.userSignOut();
+    this.props.showSignOut();
```
muffinresearch: I don't know how consistent we've been but I've been tending to call things like this handle[Whatever]. It would help to distinguish from the prop action with the same name.

public/js/reducers/management.js
```diff
@@ -20,6 +20,9 @@ export default function management(state, action) {
         });
       }
       return state;
+    case actionTypes.USER_SIGNED_OUT:
+      // Reset all view/tab state which will trigger a sign-in screen.
```
muffinresearch: :+1:  good call.

public/index.html
kumar303: Sure, I can remove direct purchase sign-in for now. We can always add it in later.

public/js/views/management.jsx
kumar303: the thing I changed it to is more like you navigate to a sign out page. For that reason I followed the pattern we already have of `showNewPage()`.

kumar303:
> If you sign in from payments-example then sign out from the management interface it doesn't work. 

Interesting! Good catch. What's happening when you click sign-out is 1) it signs you out, 2) it enters the initial app state, and 3) the initial state detects `access_token` from the query string and signs you in again. I'll fix it up.

kumar303:
> We should probably make that default to loading

Yeah, that makes more sense anyway since when it's actually signing in, the FxA modal obscures the management panel.

public/js/views/management.jsx
muffinresearch: I was unclear (highlighted the wrong line) the bit I'm talking about renaming is the hander func not the action prop.

kumar303:
Thanks for all the feedback. This new version has some changes:
* Signing out lands you on a signed out page. This makes everything a lot less complicated, especially for the access token case.
* I'll file a separate issue to get UX support for the signed out case.
* The purchase flow doesn't support direct sign-in (just like before)

PR #331, opened 2015-08-13
Update subs look and feel

muffinresearch:
mozilla_firefox

This is mainly a style + markup change to get the subs closer to the mocks. Still nothing is hooked up but the look and feel is closer. 

Note:  There's almost certainly more cleanup needed and the mobile CSS will come later. But at least the styles are closer for another pass at doing the action hookup for subs updates.

I've done this now because I needed to know what markup I'll be dealing with in order to deal with the pay method deletion which displays a subs list in some cases.

### DIFF
--- a/public/js/components/subscription.jsx
+++ b/public/js/components/subscription.jsx
@@ -1,5 +1,6 @@
 import React, { Component, PropTypes } from 'react';
 
+import PayMethodDropDown from 'components/pay-method-drop-down';
 import { gettext } from 'utils';
 import * as products from 'products';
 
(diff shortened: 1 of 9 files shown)


public/js/components/subscription.jsx
```diff
@@ -18,22 +28,29 @@ export default class Subscription extends Component {
...
-          <div>{productData.price.en}</div>
-          <div>{gettext('per month')}</div>
-        </div>
-        <nav>
-          <div className="change-pay-account">
-            <a href="#">{gettext('Change payment account')}</a>
-          </div>
-          <div className="cancel">
-            <a href="#">{gettext('Cancel subscription')}</a>
+          <img alt="" src={productData.img} width="50" />
```
tofumatt: Can you set this in CSS instead of the markup?

tofumatt:
Looks good to me, r+wc.

public/js/components/subscription.jsx
muffinresearch: Yep, I'll remove that.

PR #333, opened 2015-08-13
Add del confirmation (fixes #327, fixes #273, fixes #313)

### LINKED ISSUE #273: Show affected subscriptions on pay-method deletion

### LINKED ISSUE #313: Always load pay methods when viewing component
When you view the pay methods page (called Credit Cards & Paypal) the app should always load pay methods from the API. We have an optimization for the purchase flow that fetches pay methods at sign in which will remain in place but the management page should not rely on that cache.

### LINKED ISSUE #327: Add action to modify subs
We need an action to modify subscriptions using this endpoint: http://payments-service.readthedocs.org/en/latest/topics/api.html#post--api-braintree-subscriptions-paymethod-change-

muffinresearch:
This is rough around the edges. The actual data calls are not great because of the brittleness around iteration through subs to update them at the same time as the deletion.

That said we can definitely iterate from here.


kumar303:
Cool, I'm checking out the patch now. I'll probably leave some comments but just as a note, not for you to fix up. Have a good PTO!

public/js/actions/management.js
```diff
@@ -52,7 +59,7 @@ export function showHistory() {
 
 export function showSubscriptions() {
   return {
-    type: actionTypes.SHOW_SUBSCRIPTIONS,
+    type: actionTypes.SHOW_SUBS,
```
kumar303: I think `SUBS` is a little ambiguous

public/js/actions/subscriptions.js
```diff
@@ -33,6 +33,39 @@ export function getUserSubscriptions(fetch=api.fetch) {
...
+      method: 'get',
+      url: '/braintree/subscriptions/',
+      context: this,
+    }).then(data => {
+      console.log('got subscriptions from API:', data);
+      var filteredSubs = data.subscriptions.filter(function(item) {
+        return item.paymethod === payMethodUri;
+      });
+      dispatch({
+        type: actionTypes.GOT_FILTERED_USER_SUBS,
```
kumar303: 'filtered' is a little vague, maybe something more like `GOT_SUBS_TIED_TO_PAY_METHOD`

public/js/actions/subscriptions.js
```diff
@@ -97,3 +130,27 @@ export function createSubscription(productId, payMethod,
...
+  };
+
+  return dispatch => {
+    return fetch({
+      data: data,
+      url: '/braintree/subscriptions/paymethod/change/',
+      method: 'post',
+    }).then(() => {
+      console.log('Successfully updated subscription: ' + subscriptionUri +
+                  ' to use payMethod: ' + newPayMethodUri);
```
kumar303: this should probably dispatch something to trigger a state change

kumar303: oh, hmmm, I see it's used in the defferreds

public/js/apps/management/app.jsx
```diff
@@ -118,7 +120,18 @@ export default class ManagementApp extends Component {
           {...boundPayMethodActions}
         />
       ));
-    } else if (connector.management.view === 'SHOW_SUBSCRIPTIONS') {
+    } else if (connector.management.view === 'SHOW_CONFIRM_DEL_PAY_METHOD') {
+      children.push((
+        <ConfirmDelPayMethod
+          payMethods={connector.user.payMethods}
+          payMethodUri={connector.user.tmp.payMethodUri}
```
kumar303: when seeing this in use, it makes less sense as a user property. Maybe something tied more to the view would make sense, like `connector.management.viewData.payMethodUri`...

kumar303: that way we could wipe it out when switching tabs which is probably better than wiping it out when switching users

public/js/components/subscription.jsx
```diff
@@ -43,12 +43,12 @@ export default class Subscription extends Component {
         <div className="actions">
           <label className="vh"
             htmlFor="pm">{gettext('Change payment account')}</label>
+          {this.props.payMethods.length ?
           <PayMethodDropDown
```
kumar303: this feels like it should be indented to indicate that the logical block is still in progress

public/js/reducers/user.js
```diff
@@ -45,18 +46,30 @@ export default function user(state, action) {
...
+  if (action.type === actionTypes.GOT_USER_SUBS) {
     return Object.assign({}, initialUserState, state, {
       subscriptions: action.subscriptions,
     });
   }
 
+  if (action.type === actionTypes.GOT_FILTERED_USER_SUBS) {
+    return Object.assign({}, initialUserState, state, {
+      // This is short-lived data only relevant to the current
+      // operation.
```
kumar303: again, I think the scope of the current operation is the current management view

public/js/views/management/confirm-del-pay-method.jsx
```diff
@@ -0,0 +1,204 @@
...
+  componentDidMount() {
+    setTitle('Delete Payment Method?');
+    tracking.setPage('/confirm-del-pay-method');
+    this.props.getFilteredUserSubscriptions(this.props.payMethodUri);
+  }
+
+  handleSubmit = (e) => {
+    e.preventDefault();
+    this.setState({isSubmitting: true});
+    console.log(e);
```
kumar303: probably don't need this anymore

public/js/views/management/confirm-del-pay-method.jsx
```diff
@@ -0,0 +1,204 @@
...
+
+
+    var affectedSubs = this.props.affectedSubscriptions;
+    var deferreds = [];
+    // If there are affected subs...
+    if (affectedSubs) {
+      var select = e.target.querySelector('[name="new-pay-method"]');
```
kumar303: it looks like we could also do this by calling the filter again, like `this.getAltPayMethods()` (which could be cached)

public/js/views/management/confirm-del-pay-method.jsx
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    // Delete the card in question.
+    deferreds.push(this.props.delPayMethod(this.props.payMethodUri));
+
+    $.when(...deferreds).done(() => {
+      this.props.showPayMethods();
+    }).fail(() => {
+      this.props.error('Failed to update subscriptions');
+    });
```
kumar303: I think this will all be easier to test if it's encapsulated within the action. I know it will probably be simpler when the API changes though.

public/js/views/management/confirm-del-pay-method.jsx
```diff
@@ -0,0 +1,204 @@
...
+
+  }
+
+  getAltPayMethods = (payMethodResourceUri) => {
+    // Return the list of pay methods less the one that has
+    // a resource_uri matching `payMethodResourceUri`.
```
kumar303: nice comments everywhere in this component

public/js/views/management/confirm-del-pay-method.jsx
```diff
@@ -0,0 +1,204 @@
...
+              selectNameAttr="new-pay-method"
+              showDefaultOption={false}
+            />
+            {this.renderSubs()}
+          </div>
+        );
+      // If there's no alternative pay methods then deletion
+      // amounts to cancellation of the affected subs.
+      } else {
+        return ((
```
kumar303: double parentheses

tests/views/test.pay-methods.jsx
```diff
@@ -23,11 +23,18 @@ describe('PayMethods', function() {
       },
     ];
 
+    this.getPayMethodsStub = sinon.stub();
```
kumar303: to be consistent it would make sense to call this `fakeGetPayMethods` or `fakePayMethodGetter`

public/js/views/management/confirm-del-pay-method.jsx
muffinresearch: I realise after sleeping on this the deletion should happen only after subscription updates not at the same time. This way it only does the deletion if the subs updates worked.

kumar303:
@muffinresearch it's possible we may not integrate this before you're back because we're taking a break from management to flesh out some parts of a potential donation flow.

public/js/actions/management.js
muffinresearch: As discussed this is a commonly used abbreviation.

public/js/actions/subscriptions.js
muffinresearch: I was thinking I'll change this function and the const to something more like `getSubsByPayMethod` and `GOT_SUBS_BY_PAY_METHOD`

kumar303: +1 for `GOT_SUBS_BY_PAY_METHOD`

public/js/actions/subscriptions.js
```diff
@@ -33,6 +33,40 @@ export function getUserSubscriptions(fetch=api.fetch) {
...
+      });
+      dispatch({
+        type: actionTypes.GOT_SUBS_BY_PAY_METHOD,
+        subscriptions: filteredSubs,
+        payMethodUri: payMethodUri,
+      });
+    }).fail(apiError => {
+      console.log('error getting filtered subscriptions:',
+                  apiError.responseJSON);
+      dispatch(appActions.error('failed to get subscriptions'));
```
kumar303: maybe make this message specific to filtering by pay method as well

public/js/actions/subscriptions.js
```diff
@@ -33,6 +33,40 @@ export function getUserSubscriptions(fetch=api.fetch) {
...
+    fetch({
+      method: 'get',
+      url: '/braintree/subscriptions/',
+      context: this,
+    }, {
+      csrfToken: getState().app.csrfToken,
+    }).then(data => {
+      console.log('got subscriptions from API:', data);
+      var filteredSubs = data.subscriptions.filter(function(item) {
+        return item.paymethod === payMethodUri;
```
kumar303: could be `(item) => item.paymethod === payMethodUri` too

public/js/components/card-form.jsx
```diff
@@ -189,7 +189,7 @@ export default class CardForm extends Component {
         </div>
         <SubmitButton isDisabled={!formIsValid}
           showSpinner={this.state.isSubmitting}
-          text={this.props.submitPrompt} />
+          content={this.props.submitPrompt} />
```
kumar303: `content` seems a bit odd. What about calling the attribute `submitPrompt` ?

kumar303: hmm, well, I guess `content` is more DOM like. Meh, names.

public/js/components/subscription.jsx
```diff
@@ -43,12 +43,12 @@ export default class Subscription extends Component {
...
-            selectId="pm"
-            selectedPayMethodResource={this.props.paymethod}
-            showDefaultOption={false}
-          />
+          {this.props.payMethods.length ?
+            <PayMethodDropDown
+              payMethods={this.props.payMethods}
+              selectId="pm"
+              selectedPayMethodResource={this.props.paymethod}
+            /> : <p className="warn">{gettext('No pay methods available')}</p>}
```
kumar303: Since this is user facing I think it should say 'No payment methods...'

public/js/reducers/management.js
```diff
@@ -26,6 +27,25 @@ export default function management(state, action) {
...
+          payMethodUri: action.payMethodUri,
+        },
+      });
+    case actionTypes.SHOW_CONFIRM_DEL_PAY_METHOD:
+      return Object.assign({}, initialMgmtState, state, {
+        // Ephemeral data for this operation only.
+        tab: 'pay-methods',
+        view: action.type,
+        viewData: {
+          operation: actionTypes.SHOW_CONFIRM_DEL_PAY_METHOD,
```
kumar303: I don't see `operation` used anywhere. Is it still needed? Isn't the operation already defined as `view` ?

public/js/reducers/user.js
```diff
@@ -75,5 +75,15 @@ export default function user(state, action) {
     });
   }
 
+  if (action.type === actionTypes.SHOW_CONFIRM_DEL_PAY_METHOD) {
+    return Object.assign({}, initialUserState, state, {
+      // Ephemeral data for this operation only.
+      tmp: {
+        operation: actionTypes.SHOW_CONFIRM_DEL_PAY_METHOD,
```
kumar303: same question here about `operation`

public/js/utils.jsx
```diff
@@ -87,3 +85,19 @@ export function parseQuery(url){
 
   return data;
 }
+
+
+/**
+ * Given a list of pay-method objects return one that matches
+ * the payMethodUri. If no matches returns null.
```
kumar303: last sentence could use some re-wording, maybe : 'If none match, it returns null'

public/js/utils.jsx
```diff
@@ -87,3 +85,19 @@ export function parseQuery(url){
...
+ * @param {list} payMethods - a list of pay-method objects.
```
kumar303: list not array? I don't know how the docstrings work

public/js/views/management/confirm-del-pay-method.jsx
```diff
@@ -0,0 +1,220 @@
...
+        }).fail(() => {
+          // Abort any outstanding requests
+          console.log('Aborting outstanding requests');
+          for (var j = 0; j < deferreds.length; j += 1) {
+            deferreds[j].abort();
+          }
+          this.props.error('Failed to update subscriptions');
+        });
+      } else {
+        // What we're doing amounts to cancellation.
```
kumar303: not sure you need this comment since the log message says it

public/js/views/management/confirm-del-pay-method.jsx
```diff
@@ -0,0 +1,220 @@
...
+      } else {
+        return (
+          <div>
+            <PayMethodItem
+              inputType="hidden"
+              payMethod={this.state.payMethod}
+            />
+            <hr />
+            <p className="cancellation"
+              >{gettext('These subscriptions will be cancelled')}</p>
```
kumar303: nit: dangling angle bracket looks odd to me

tests/views/test.del-pay-method-confirmation.jsx
```diff
@@ -0,0 +1,199 @@
...
+          'public_id': 'mozilla-concrete-mortar',
+          'resource_pk': 1,
+          'resource_uri': '/generic/product/1/',
+        },
+      }, {
+        'id': 2,
+        'paymethod': '/pay-method/2/',
+        'resource_uri': '/sub/2/',
+        'seller_product': {
+          'external_id': 'mozilla-concrete-brick',
```
kumar303: you can probably leave off `external_id` because I don't think we use it anywhere in the whole app

tests/views/test.del-pay-method-confirmation.jsx
```diff
@@ -0,0 +1,199 @@
...
+
+    this.delPayMethodStub = sinon.stub();
+    this.delPayMethodStub.returns($.Deferred().resolve());
+
+    this.getSubsByPayMethodStub = sinon.stub();
+    this.showDelPayMethodStub = sinon.stub();
+    this.showPayMethodsStub = sinon.stub();
+    this.updateSubPayMethodStub = sinon.stub();
+
+    this.mountView = function(payMethodUri,
```
kumar303: I think it makes the calling code a lot more flexible and easier to read when you use keyword args like:
````
this.mountView = function({payMethodUri, 
                           affectedSubscriptions=this.fakeSubs,
                           payMethods=this.fakePayMethods})
this.mountView({payMethodUri: '/pay-method/2/'})
````

File: public/js/actions/subscriptions.js [hunk 25]
```diff
@@ -33,6 +33,40 @@ export function getUserSubscriptions(fetch=api.fetch) {
 }
 
 
+export function getSubsByPayMethod(payMethodUri, fetch=api.fetch) {
```
kumar303: I didn't see a direct test for this action and I think it needs one so that the Ajax logic is covered

kumar303:
I would like to see a direct test for the `getSubsByPayMethod` action but other than that, r+wc

public/js/reducers/management.js
muffinresearch: Yeah, I'd missed one having ripped that out of the other places it was used already.

tests/views/test.del-pay-method-confirmation.jsx
muffinresearch: Yeah, good call

PR #335, opened 2015-08-14
preserve tab on error fixes #329

### LINKED ISSUE #329: Preserve tab/view when card submission errors are handled
in the management tab, I noticed this https://github.com/kumar303/payments-ui/blob/sign-out/public/js/reducers/management.js#L67

````
    case actionTypes.CREDIT_CARD_SUBMISSION_ERRORS:
      return Object.assign({}, initialMgmtState, {
        cardSubmissionErrors: action.apiErrorResult,
      });
````
this is wrong, it should preserve the tab state so that the page focus isn't lost.

### PR DESCRIPTION
**Pull request by andymckay**
Also fix ordering to keep the linter happy

### DIFF
--- a/public/js/reducers/management.js
+++ b/public/js/reducers/management.js
@@ -67,6 +67,7 @@ export default function management(state, action) {
     case actionTypes.CREDIT_CARD_SUBMISSION_ERRORS:
       return Object.assign({}, initialMgmtState, {
         cardSubmissionErrors: action.apiErrorResult,
+        tab: state.tab,
       });
     case actionTypes.CLOSE_MODAL:
       return initialMgmtState;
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: public/js/reducers/management.js [hunk 1]
```diff
@@ -67,6 +67,7 @@ export default function management(state, action) {
     case actionTypes.CREDIT_CARD_SUBMISSION_ERRORS:
       return Object.assign({}, initialMgmtState, {
```
kumar303: I think there could be more things besides `tab` that we'd want to preserve so it should just inherit from the current state like this:
````
return Object.assign({}, state, {
  cardSubmissionErrors: action.apiErrorResult,
});
````

kumar303:
r+wc

kumar303:
btw, I've noticed that putting the issue number in the commit message creates a lot of noise in the issue thread since it will add a link for every rebase (if you choose to rebase). Adding it to the title of the PR only adds one link back to the PR.

PR #337, opened 2015-08-14
localize product name

### LINKED ISSUE #307: Show localized seller name in transaction history
If a seller name is localized for the current locale, show it. Otherwise fall back to the default localization. History view landed in https://github.com/mozilla/payments-ui/pull/303/

andymckay:
* not sure how we are doing front end localisation, but a pull request is a good way to find out...
* provides a generic configGetText that can work on any localised object in payments-configuration
* would be easier if it was a JS module, i guess? :wink: 
* fixes #307

### DIFF
--- a/public/js/utils.jsx
+++ b/public/js/utils.jsx
@@ -17,6 +17,23 @@ export function defaults(object, opt) {
   return object;
 }
 
+/**
+* Similar to getText, but works on a product JSON object returned by
+* the payments configuration so we can figure out the best data to show.
+*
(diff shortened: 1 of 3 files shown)


tofumatt:
Looks good to me, what about tests for locales like en-US versus en-GB versus en?

andymckay:
we haven't really got a plan for localisation fall back yet, still not sure what we are doing on this tbh

tofumatt:
Makes sense to address later (outside the scope of this PR). :-)

Issue #334, opened 2015-08-13
Create a better signed-out view


kumar303:
We decided we need a signed-out view in the management app. See it in action here (gif screencast)  https://github.com/mozilla/payments-ui/pull/328 The user will only see this if they click the sign-out button. That is, when the page first loads they won't see it; they will be prompted to sign in automatically.

Bram, can you help us figure out the best UX for this?

Here are its key points:
* It needs to let the user sign back in
* It needs to hide links to all other pages
* This will serve as the background for the FxA sign-in overlay (which we don't really have control over)

brampitoyo:
What do you think about using either one of these four options? (I like the two on the right-hand side the best).

![ux-management-wireframe-signed-out](https://cloud.githubusercontent.com/assets/51007/9266365/204b7868-4294-11e5-8160-340f1fb90891.png)

kumar303:
Cool. I like the one in the upper right the best -- the dark background with the centered sign-in. I think that will work best because the FxA sign-in overlay is white and it's also centered.

brampitoyo:
The only objection I have against that design is that the transition between the dark background of the sign-in page and the mostly-light background of the management interface will be jarring, but I think it’s a very minor issue.

kumar303:
The bottom right (light background) would work fine too. I think the ones that retain the lefthand sidebar would be odd because the FxA overlay will be centered.

kumar303:
I'll take a stab at it

kumar303:
@brampitoyo I went with the lighter background one but funny enough, the FxA overlay dims the background and we don't really have control over that. You can see it in action here: https://github.com/mozilla/payments-ui/pull/341

kumar303:
I merged in the light background one. After about 10 minutes, it should be visible on the dev site: http://pay.dev.mozaws.net:8000/management.html

PR #341, opened 2015-08-17
Re-style the sign-out view (#334)

kumar303:
NOTE: there is some glitch in my screencast converter where you see a ghost of a sidebar during the sign-in spinner. This does not appear in reality.

sign-out mov

Supports https://github.com/mozilla/payments-ui/issues/334


andymckay:
r+

Issue #338, opened 2015-08-17
Start a donation payment without signing in


kumar303:
If the product being purchased is a donation (not a subscription) then user sign-in should be optional. If they choose to sign-in, they may get access to their saved payment methods but it's not mandatory.

Supports https://github.com/mozilla/payments/issues/117

PR #343, opened 2015-08-18
Start a donation without sign-in (#338)

kumar303:
Supports https://github.com/mozilla/payments-ui/issues/338

### DIFF
--- a/public/js/apps/transaction/app.jsx
+++ b/public/js/apps/transaction/app.jsx
@@ -12,6 +12,7 @@ import DefaultTransaction from 'views/transaction';
 
 import * as appActions from 'actions/app';
 import * as userActions from 'actions/user';
+import * as products from 'products';
 import { parseQuery } from 'utils';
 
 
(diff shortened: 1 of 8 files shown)


public/js/apps/transaction/app.jsx
```diff
@@ -50,6 +51,10 @@ export default class TransactionApp extends Component {
     var state = this.state;
     var SignIn = this.props.SignIn;
     var Transaction = this.props.Transaction;
+    var product = products.get(state.productId);
+    var signInRequired = product.seller.kind !== 'donations';
+    console.log('sign-in required to transact product?', state.productId,
+                signInRequired ? 'Yes': 'No');
```
andymckay: this looks cool, i think we'll need to think about making explicit authentication restrictions in payments-config and not infer it from anything to cope with non-authenticated subscriptions. seperate bug though...

andymckay:
looks like a couple of eslint issues

andymckay:
r+with travis fixes

public/js/apps/transaction/app.jsx
kumar303: I think of this still as a strawman type of patch. A few things we can do later:

- make sign-in optional for donations anyway
- maybe rely on the API to see if we need to sign-in: try to make a request and see if we get a 403 before we try to sign in

PR #345, opened 2015-08-18
make per month depend on the config

andymckay:
* fixes mozilla/payments-env#51 which I totally filed in the wrong repo, oops.

### DIFF
--- a/public/js/components/product-detail.jsx
+++ b/public/js/components/product-detail.jsx
@@ -11,13 +11,15 @@ export default class ProductDetail extends Component {
   render() {
 
     var productData = products.get(this.props.productId);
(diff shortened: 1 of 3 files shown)


public/js/components/product-detail.jsx
```diff
@@ -11,13 +11,15 @@ export default class ProductDetail extends Component {
   render() {
 
     var productData = products.get(this.props.productId);
+    var recurrence = (productData.recurrence === 'monthly' ?
+      <div className="recurrence">{gettext('per month')}</div> : '');
```
kumar303: looks good for now but will need fixing up for 'per year' and others

kumar303:
r+

Issue #344, opened 2015-08-18
Set up the CSRF token when processing a braintree token


kumar303:
Because donations don't require sign-in, we have to set up the CSRF somewhere. The current plan is to do it after posting to the braintree token generator, which gives us a CSRF token now: http://payments-service.readthedocs.org/en/latest/topics/api.html#post--api-braintree-token-generate-

Supports https://github.com/mozilla/payments/issues/117

PR #346, opened 2015-08-19
Store CSRF token after generating braintree token (#344)

kumar303:
This also fixes the way we used to handle CSRF data. The old way was to use a setter and rely on a JS module var. The new way uses the redux model for keeping CSRF state all in one place.

Supports: https://github.com/mozilla/payments-ui/issues/344


andymckay:
not really sure i follow whats going on, and my eyes started to glaze over a bit towards the bottom - but nothing that seemed a problem

r+ with sort of worries that i don't know what i'm really r?ing

kumar303:
Yeah, it's a lot of refactoring, sorry. The crux of the patch is moving to a redux style of data handling.

Pros:
- data flow is more predictable
- csrf handling becomes part of the redux capture/replay stack for debugging, monitoring, and error reporting, etc
- csrf was the last piece of data in the app which wasn't reduxified, I think

Cons:
- all calls to api.fetch() must provide an explicit token. I couldn't think of a good way to DRY this up but maybe there is a way.

Maybe @muffinresearch can also take a look when he's back.

Issue #349, opened 2015-08-19
Payments failing


andymckay-limited-access:
I'm betting this is related to recent donations changes so giving that label.

On starting a payment, after completing the FxA login, I get an error in the console: "Invalid product: undefined".

https://www.dropbox.com/s/msk88s98szazvac/Screenshot%202015-08-19%2016.47.06.png?dl=0

Reproducible locally and on http://pay.dev.mozaws.net/.

andymckay-limited-access:
Is the result of # appearing in the URL? 

http://pay.dev:8000/?access_token=...&product=mozilla-concrete-brick fails
http://pay.dev:8000/#/?access_token=...&product=mozilla-concrete-brick works

andymckay-limited-access:
Bisect says its https://github.com/mozilla/payments-ui/commit/e5e8b1cf3c81bb2eadce4e4d325d68f27a512f7d

andymckay-limited-access:
The # causes this split to ignore everything else: https://github.com/mozilla/payments-ui/blob/master/public/js/utils.jsx#L75

kumar303:
Aha! Good spotting, sorry about that. I was getting this error too but I didn't have time to pin it down.

r? https://github.com/mozilla/payments-ui/pull/350

andymckay-limited-access:
This is caused by this change. Looks like there's a good reason for it. https://github.com/mozilla/payments-client/commit/0e4905471d5afa2f1f8653c66973c174c68efced

PR #350, opened 2015-08-20
Fix parsing queries in URLs that have anchors (#349)

kumar303:
Fixes https://github.com/mozilla/payments-ui/issues/349

### DIFF
--- a/public/js/utils.jsx
+++ b/public/js/utils.jsx
@@ -72,12 +72,14 @@ export function parseQuery(url){
   // @param {string} url - complete URL that may or may not include a query
   // @returns {object} - parsed query string parameter names mapped to values
   //
-  var baseUrl = url.split('#')[0];
-  var urlParts = baseUrl.split('?');
+  var urlParts = url.split('?');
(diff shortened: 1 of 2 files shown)


andymckay:
I'd be ok with this to workaround the problem, but I'm curious why the # exists in the first place, seems like something we shouldn't be adding?

kumar303:
What happens is we have some a tags that link to an href of `#` and then react handles the event instead. This results in URLs like `http.../?access_token=foo#` which I think is expected behavior. In other words, I think the query parser is doing the right thing to handle trailing anchors and strip them out but correct me if I'm wrong. 

I'm not totally clear why `#` is showing up in the URL at all because our event handlers all do `e.preventDefault()` so maybe that is something else to look into. Regardless, anyone could append an anchor to the query string in the URL so the parser should handle that.

andymckay:
Yeah having a # at the end is cool, but at the beginning of the URL. Anyways without this we are hosed, so let's merge this, but I'll keep digging on the real cause.

PR #351, opened 2015-08-20
skip email on completion page

### LINKED ISSUE #340: Payment completion page for anonymous payments
For anonymous payments (such as donations), we may not have an email address so the payment completion page needs to be a little different.

Supports: https://github.com/mozilla/payments/issues/117

### LINKED ISSUE #348: Fix up other uses of per month
In the receipts and in the management interface.

They should be checking the product config and probably make a little utility function out of that.

andymckay:
* fixes #340 
* I feel like doing these if's might not scale, but also breaking everything down into smaller and smaller parts seem a little crazy
* seems to work, but guidance welcome

### DIFF
--- a/public/js/views/transaction/complete-payment.jsx
+++ b/public/js/views/transaction/complete-payment.jsx
@@ -42,14 +42,17 @@ export default class CompletePayment extends Component {
   }
 
   render() {
+    var email = (this.props.userEmail ?
+      <p className="receipt">
+        {gettext('Your receipt has been sent to')}
+        <span className="email">{this.props.userEmail}</span>
+      </p> : '');
+
     return (
       <div className="complete">
         <ProductDetail productId={this.props.productId} />
         <p className="accepted">{gettext('Payment Accepted')}</p>
-        <p className="receipt">
-          {gettext('Your receipt has been sent to')}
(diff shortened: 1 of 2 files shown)


public/js/views/transaction/complete-payment.jsx
```diff
@@ -42,16 +42,19 @@ export default class CompletePayment extends Component {
...
+      </p> : '');
+
     return (
       <div className="complete">
         <ProductDetail productId={this.props.productId} />
         <p className="accepted">{gettext('Payment Accepted')}</p>
-        <p className="receipt">
-          {gettext('Your receipt has been sent to')}
-          <span className="email">{this.props.userEmail}</span>
-        </p>
```
kumar303: I think `{email}` should go here, not below the submit button to follow the mocks

kumar303:
> I feel like doing these if's might not scale

I actually like this approach because it's very direct: it only checks for the presence of email which is the right kind of if statement

public/js/views/transaction/complete-payment.jsx
andymckay: oh oops, thanks

kumar303:
r+

Issue #339, opened 2015-08-17
Execute one-off payment transaction


kumar303:
When paying for a product that is a one-off payment (such as a donation), execute the transaction correctly. The endpoint will be different and the button will say something different.

Supports https://github.com/mozilla/payments/issues/117

PR #352, opened 2015-08-20
Process a one-time sale (like donations) (#339)

kumar303:
Fixes https://github.com/mozilla/payments-ui/issues/339

Requires the backend changes from https://github.com/mozilla/payments-service/issues/134


public/js/apps/transaction/app.jsx
```diff
@@ -37,6 +37,8 @@ export default class TransactionApp extends Component {
     this.state = {
       accessToken: qs.access_token,
       productId: qs.product,
+      // This is an amount to pay, which only applies to donations.
```
andymckay: I predict that comment will rot, its possible we'll have a "pay what you want" product in the future. Don't want to hang too many things off of one settings value...

public/js/components/product-detail.jsx
```diff
@@ -13,12 +16,23 @@ export default class ProductDetail extends Component {
     var productData = products.get(this.props.productId);
     var recurrence = (productData.recurrence === 'monthly' ?
       <div className="recurrence">{gettext('per month')}</div> : '');
+    var price;
+
+    // TODO: localize/format prices with currency symbol.
+    if (productData.price && Object.keys(productData.price).length) {
+      console.log('Showing configured price for product', productData.id);
+      price = productData.price.en;
```
andymckay: oh nice

tests/actions/test.subs-actions.js
```diff
@@ -70,47 +70,34 @@ describe('subscription actions', function() {
...
-                                payMethod=fakeCard) {
-      var deferredAction = subActions.createSubscription(
-                                          'product-id',
-                                          payMethod,
-                                          'braintree-token',
-                                          fetch,
-                                          client);
+    function createSubscription({fetch=fakeFetch,
+                                 client=helpers.FakeBraintreeClient,
+                                 payMethod=fakeCard} = {}) {
```
andymckay: Need to a find a doc on whats going on with this function args pattern. Got a link?

tests/actions/test.transaction.js
```diff
@@ -99,4 +106,131 @@ describe('transactionActions', function() {
...
+
+    it('should create a subscription for recurring products', function() {
+      var fetch = helpers.fakeFetch();
+      processPayment({fetch: fetch, productId: 'mozilla-concrete-brick'});
+      assert.equal(fetch.firstCall.args[0].url, '/braintree/subscriptions/');
+    });
+
+    it('should create a sale for one-off products', function() {
+      var fetch = helpers.fakeFetch();
+      processPayment({fetch: fetch, productId: 'mozilla-foundation-donation'});
```
andymckay: For a later bug, but i'm not keen on test in payments-ui being dependent upon data inside payments-config. Ideally we can define a new set of values for the config for these tests.

andymckay:
r+ man i'll be glad when @muffinresearch is back :)

public/js/apps/transaction/app.jsx
kumar303: oh yeah, true

tests/actions/test.subs-actions.js
kumar303: named parameters: http://www.2ality.com/2011/11/keyword-parameters.html

There's actually one thing you can do in JS that you can't do in Python! You can invoke the default by passing in undefined:
````
function foo({param='default-value'}) {
  console.log(param);
}
foo({param: undefined});  // prints 'default-value'
````
In python, you have to leave off the parameter altogether to accomplish this. It's not a big deal but sometimes it leads to silly code like this:

````
def foo(param='the-default'):
    print param

if thing:
    foo(param=thing)
else:
    foo()
````
or like this:

````
def foo(param=None):
    if not param:
        param = 'the-default'
    print param

foo(param=thing)
````

tests/actions/test.transaction.js
kumar303: Yeah, I was thinking about that too. It's a bad pattern that we keep building on. Filed: https://github.com/mozilla/payments-ui/issues/353

It's a little tricky for a couple reasons:
- lots of code uses `products.get(id)` rather than accept a complete object
- the config is in flux so it's kind of nice to use real-ish configured products

tests/actions/test.subs-actions.js
andymckay: Thanks for the link, thats what i was looking for, specifically the {...} = {}. Yay for learning in pull requests.

kumar303:
This needs https://github.com/mozilla/payments-service/issues/134 but I'm going to get it merged in anyway since it won't break any existing stuff.

Issue #97, opened 2015-06-11
Localize currency display for subscription price better


kumar303:
We want to show "$9.99 per month" in the subscription purchase page. This landed in https://github.com/mozilla/payments-ui/pull/94/ but we can localize it better. There may be some prior art in [Zamboni](https://github.com/mozilla/zamboni) for how we do things like this. The decimal formatting may be different too, for example: "€0,99 per month"

brampitoyo:
One thing we could do to avoid the comma-in-cent problem (ie. whether a cent should be preceded by a period or a coma) is to separate the cent value from the nominal price (we will also have to separate the currency symbol), like so:

![price-currencies](https://cloud.githubusercontent.com/assets/51007/8123352/2273ae5c-111c-11e5-95dd-95ba0cfddf53.png)

Out of concern for space, I put “per month” value directly below the price, but this means that we can easily take that line out for one-off purchases.

kumar303:
oh yeah, cool, that seems like it would work. There could be some tricky things with Yen, not sure. Those price numbers get really high. Anyway, as far as I know, these mocks will work.

andymckay:
That's cool. I do think that might require a bit of thinking about. Here's all the number formats in Babel for $10 USD, per locale.

```
    "en": "$10.00", 
    "ca": "US$10,00", 
    "it": "US$\u00a010,00", 
    "cs": "10,00\u00a0US$", 
    "cy": "US$10.00", 
    "eu": "10,00\u00a0US$", 
    "et": "10,00\u00a0$", 
    "id": "US$10,00", 
    "es": "10,00\u00a0US$", 
    "ru": "10,00\u00a0$", 
    "nl": "US$\u00a010,00", 
    "pt": "US$10,00", 
    "tr": "10,00\u00a0$", 
    "es_CL": "US$10,00", 
    "lt": "10,00\u00a0$", 
    "pa": "US$\u00a010.00", 
    "es_AR": "10,00\u00a0US$", 
    "rm": "10.00\u00a0US$", 
    "pl": "10,00\u00a0US$", 
    "fr": "10,00\u00a0$US", 
    "zh_TW": "US$10.00", 
    "pt_BR": "US$10,00", 
    "de": "10,00\u00a0$", 
    "da": "10,00\u00a0$", 
    "ff": "10,00\u00a0US$", 
    "hu": "10,00\u00a0$", 
    "ja": "$10.00", 
    "he": "10.00\u00a0US$", 
[...]

PR #356, opened 2015-08-21
Show variable price in USD for now

kumar303:
This will be expanded in https://github.com/mozilla/payments-ui/issues/97


andymckay:
r+ thanks

PR #363, opened 2015-08-23
Rename grunt tasks to be more descriptive (fixes #361)

### LINKED ISSUE #361: Differentiate grunt serve and grunt start
Aaargh, I can never remember what does what and get these two confused. Could we call grunt serve something like grunt hot-webpack-reloading-action or something?

tofumatt:
Old tasks names kept around for now, until we update payments-env. (Assuming this is an okay change.)

These names make more sense to me, not sure what others think.

Closes #361.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -38,12 +38,29 @@ module.exports = function(grunt) {
 
   grunt.registerTask('build', ['clean:dist', 'sass', 'webpack']);
 
-  grunt.registerTask('serve',
+  // The dev server used for live reloading in the browser.
+  grunt.registerTask('serve', 'Dev server with webpack hot module reloading',
     ['clean:dist', 'sass:dev', 'webpack-dev-server:start', 'watch:sass']);
-  grunt.registerTask('start',
-    ['clean:dist', 'sass:dev', 'webpack:dev', 'watch:sass']);
-  grunt.registerTask('start-email',
+  // Used to serve email CSS.
+  grunt.registerTask('watch-email',
     ['clean:dist', 'sass:email', 'webpack:dev', 'watch:sassemail']);
+  // Regular grunt server, used when running inside the payments-env docker.
+  grunt.registerTask('watch-static',
+    ['clean:dist', 'sass:dev', 'webpack:dev', 'watch:sass']);
+  // Legacy tasks still used by payments-env, see:
+  // https://github.com/mozilla/payments-ui/issues/361
+  grunt.registerTask('start', 'Deprecated; use `watch-static` instead',
+                     function() {
+    grunt.log.writeln(
+      '"grunt start" is deprecated; use `grunt watch-static` instead');
+    grunt.task.run('watch-static');
+  });
+  grunt.registerTask('start-email', 'Deprecated; use `watch-email` instead',
+                     function() {
+    grunt.log.writeln(
+      '"grunt start-email" is deprecated; use `grunt watch-email` instead');
+    grunt.task.run('watch-email');
+  });
 
   grunt.registerTask('test', ['karma:run', 'eslint', 'csslint:lax']);
 
--- a/README.md
+++ b/README.md
@@ -11,6 +11,20 @@
 This project comprises all styles, behaviour and interfaces for
 [mozilla/payments](https://github.com/mozilla/payments).
 
+
+## Grunt tasks
+
+Run tasks with `grunt` e.g. `grunt watch-static` or `grunt serve`.
+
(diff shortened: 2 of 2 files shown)


Gruntfile.js
```diff
@@ -38,12 +38,26 @@ module.exports = function(grunt) {
...
   grunt.registerTask('build', ['clean:dist', 'sass', 'webpack']);
 
+  // Regular grunt server, used when running inside the payments-env docker.
   grunt.registerTask('serve',
-    ['clean:dist', 'sass:dev', 'webpack-dev-server:start', 'watch:sass']);
-  grunt.registerTask('start',
     ['clean:dist', 'sass:dev', 'webpack:dev', 'watch:sass']);
-  grunt.registerTask('start-email',
+  // The dev server used for live reloading in the browser.
+  grunt.registerTask('serve:dev',
```
muffinresearch: This is a bit ambiguous as 'mytask:bar' would normally mean the bar target of the mytask task [1]

[1] http://gruntjs.com/creating-tasks#multi-tasks

muffinresearch: Using `serve` for the standard case and email is a bit odd because nothing is served by grunt wheres in the webpack dev-server case it is actually serving something on a port.

I think `start` vs `serve` are already clearly delineated - maybe the issue here is more docs. 

`watch` might make more sense than `start` but watch is already a task so you can't use `grunt watch` / `grunt watch-email`. We could do `watch-static` and `watch-email` though.

Maybe we should do better docs in a table on the README so it's super clear?

E.g:

## Grunt tasks

Each task is run with the grunt command e.g. `grunt watch-static`.

| Task name       | Description |
| ------------ | --------------- | 
| serve  | Watching the CSS/JS for changes and serve the static files via webpack dev server to enable hot module reloading of JS |
| watch-static | watch the files for changes and update CSS/JS on changes |
| watch-email | watch the email specific static files for changes and update CSS/JS on changes |

tofumatt: Aha, I didn't realise that syntax was for multi tasks, my bad.

`watch` feels a *lot* better than `start`, I would go for that over my suggestions.

I also like the tables in the README idea. I'll fix this PR up to reflect these ideas and push a new version.

muffinresearch: Cool, give me a shout if you want me to take a look.

tofumatt:
Sorry, finally updated this.

I used `watch` over `start`, as `start` implies the "default" task to run for the package (npm packages will typically define `npm start` to be `grunt serve`). "watch" implies a server watching for changes, so I stuck with that.

r? @kumar303 or @muffinresearch

Gruntfile.js
```diff
@@ -38,12 +38,27 @@ module.exports = function(grunt) {
...
     ['clean:dist', 'sass:email', 'webpack:dev', 'watch:sassemail']);
+  // Regular grunt server, used when running inside the payments-env docker.
+  grunt.registerTask('watch-static',
+    ['clean:dist', 'sass:dev', 'webpack:dev', 'watch:sass']);
+  // Legacy tasks still used by payments-env, see:
+  // https://github.com/mozilla/payments-ui/issues/361
+  grunt.registerTask('start', function() {
+    grunt.log.writeln(
+      '"grunt start" is deprecated; use `grunt watch-static` instead');
+    grunt.task.run('serve');
```
muffinresearch: This should run watch-static?

Gruntfile.js
```diff
@@ -38,12 +38,27 @@ module.exports = function(grunt) {
...
+  });
+  grunt.registerTask('start-email', function() {
+    grunt.log.writeln(
+      '"grunt start-email" is deprecated; use `grunt watch-email` instead');
+    grunt.task.run('serve:email');
```
muffinresearch: This should be watch-email

Gruntfile.js
muffinresearch: When using funcs in registerTask the middle arg between the taskname  + func can be a description output by --help. See http://gruntjs.com/api/grunt.task#grunt.task.registertask for more details. That might be nice for showing what's deprecated.

muffinresearch:
r+wc

kumar303:
Thanks! Looks good to me

Gruntfile.js
tofumatt: Oops, I was obviously too hungry when I looked this over, yeah it should be `watch-static`. And I'll look at the `--help` bit too.

tofumatt:
Cool, will merge after the fixes, cheers!

PR #365, opened 2015-08-25
Prefix private action helpers with underscore (fixes #364)

### LINKED ISSUE #364: prefix private action helpers with underscores
There are a couple action helpers that should not ever be bound and called directly by event handlers. They are `actions/subscriptions:createSubscription` and `actions/transactions:processOneTimePayment`. We were thinking it would be good to prefix those with underscores so that no one is tempted to try and use them directly.

muffinresearch:


kumar303:
thanks! r+

PR #367, opened 2015-08-26
Fix completion button text (fixes #366)

### LINKED ISSUE #366: Completion button says "submit" not "Ok"
This is due to a change where the prop for the submitbutton is content now not text as it can be a string or nodes.

muffinresearch:


tests/components/test.product-detail.jsx
```diff
@@ -28,9 +28,10 @@ describe('ProductDetail', function() {
   });
 
   it('should not show recurrence for none', function() {
-    var view = mountView({productId: "mozilla-foundation-donation"});
-    expect(() => helpers.findByClass(view, 'recurrence'))
-      .to.throw('Did not find exactly one match');
+    var view = mountView({productId: 'mozilla-foundation-donation'});
+    assert.throws(
+      () => helpers.findByClass(view, 'recurrence'),
+      Error, /Did not find exactly one match/);
```
kumar303: any particular reason to remove expect?

muffinresearch: As discussed in the stand-up expect hasn't been being used so this is more about consistency as we're sticking to assert style tests rather than a mix.

kumar303:
r+

PR #373, opened 2015-08-28
Add email field when required (fixes #358)

### LINKED ISSUE #358: Add email into the payment flow
Add in email into the payment flow if its required as per the config. @brampitoyo had this in earlier mocks, so we can probably just use that for now.

Supports: https://github.com/mozilla/payments/issues/117

muffinresearch:
This displays the email field in the flow when required and does the necessary client-side validation + feedback as you type.

If passes the email as far as https://github.com/mozilla/payments-ui/blob/master/public/js/actions/transaction.js#L66 after that it will need additional changes to pass it to the API.

### DIFF
--- a/public/js/components/card-form.jsx
+++ b/public/js/components/card-form.jsx
@@ -3,7 +3,7 @@ import React, { Component, PropTypes } from 'react';
 
 import CardInput from 'components/card-input';
 import SubmitButton from 'components/submit-button';
-import { gettext } from 'utils';
+import { gettext, validateEmailAsYouType } from 'utils';
 
 
 const defaultFieldAttrs = {
@@ -41,6 +41,19 @@ const fieldProps = {
     'placeholder': gettext('Card number'),
     'validator': CardValidator.number,
   },
+  email: {
+    'attrs': {
+      'type': 'email',
+    },
+    'classNames': ['email'],
+    'errors': {
+      invalid: gettext('Invalid email address'),
+    },
+    'id': 'email',
+    'pattern': null,
+    'placeholder': gettext('Email address'),
+    'validator': validateEmailAsYouType,
+  },
   expiration: {
     'attrs': defaultFieldAttrs,
     'classNames': ['expiration'],
@@ -72,6 +85,7 @@ export default class CardForm extends Component {
   static propTypes = {
     card: PropTypes.object,
     cvv: PropTypes.object,
+    emailFieldRequired: PropTypes.bool,
     expiration: PropTypes.object,
     handleCardSubmit: PropTypes.func.isRequired,
     id: PropTypes.string.isRequired,
@@ -80,8 +94,8 @@ export default class CardForm extends Component {
   }
 
   static defaultProps = {
-    // This should always be overidden with a localized value.
-    submitPrompt: 'Submit',
+    emailFieldRequired: false,
+    submitPrompt: gettext('Submit'),
   }
 
   constructor(props) {
@@ -90,6 +104,7 @@ export default class CardForm extends Component {
       card: '',
       cardType: null,
       cvv: '',
+      email: '',
       errors: {},
       expiration: '',
       isSubmitting: false,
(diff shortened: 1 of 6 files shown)


public/js/components/card-form.jsx
```diff
@@ -123,11 +138,16 @@ export default class CardForm extends Component {
...
     this.setState({isSubmitting: true});
-    this.props.handleCardSubmit({
+    var formData = {
       number: this.state.card,
       cvv: this.state.cvv,
       expiration: this.state.expiration,
-    });
+    };
+    if (this.props.emailFieldRequired) {
+      this.props.handleCardSubmit(formData, this.state.email);
```
kumar303: since email is optional I think it would be better to pass it in as a keyword:
````
handleCardSubmit(..., {email: this.state.email});
````

File: public/js/components/card-input.jsx [hunk 2]
```diff
@@ -98,14 +98,23 @@ export default class CardInput extends Component {
...
         { showCardIcon ?
           <PayMethodIcon payMethodType={this.props.cardType} /> : null }
-        <MaskedInput
-          {...this.props.attrs}
-          id={this.props.id}
-          className={this.props.id + '-input'}
-          onChange={this.props.onChangeHandler}
-          pattern={pattern}
-          placeholder={placeholder}
-        />
```
kumar303: I guess this could also be:
````
var extraProps = {};
if (pattern) {
  extraProps.pattern = pattern;
}
<MaskedInput
            {...this.props.attrs}
            {...extraProps}
            id={this.props.id}
            className={this.props.id + '-input'}
            onChange={this.props.onChangeHandler}
            placeholder={placeholder}
          />
````

that way you wouldn't have to repeat so many fields.

Comment (kumar303) [inline public/js/components/card-input.jsx]: oh nevermind, I see that one is a regular input and another is MaskedInput. I guess copying the attributes is fine.

File: public/js/utils.jsx [hunk 3]
```diff
@@ -101,3 +101,60 @@ export function getPayMethodFromUri(payMethods, payMethodUri) {
...
+  var parts = email.split('@');
+  var localLength = parts[0] && parts[0].length;
+  var domainLength = parts[1] && parts[1].length;
+
+  // Original regexp from:
+  //  http://blog.gerv.net/2011/05/html5_email_address_regexp/
+  // Modified to remove the length checks, which are done later.
+  // IETF spec: http://tools.ietf.org/html/rfc5321#section-4.5.3.1.1
+  // NOTE: this does *NOT* allow internationalized domain names.
+  /* eslint-disable max-len */
```
kumar303: a way to get around linting is to do: `new RegExp('...' + '...' + '...')` which might actually be nice since it would split up the regex into smaller readable chunks.

tests/components/test.card-form.jsx
```diff
@@ -84,6 +85,21 @@ describe('Card Form', function() {
     assert.ok(TestUtils.isCompositeComponent(expirationError));
   });
 
+ it('shows a email error on invalid input', function() {
```
kumar303: nit: "shows an email...."

tests/components/test.card-form.jsx
```diff
@@ -84,6 +85,21 @@ describe('Card Form', function() {
...
+    var view = mountView({emailFieldRequired: true});
+    view.handleChange({
+      target: {
+        value: 'foo@@@@@',
+        id: 'email',
+      },
+    });
+    console.log(React.findDOMNode(view));
```
kumar303: still needed?

tests/test.utils.jsx
```diff
@@ -120,3 +120,48 @@ describe('utils.configGetText', function() {
...
+
+  it('should be invalid', function() {
+    assert.notOk(utils.isValidEmail('foo@'));
+  });
+
+  it('should be valid', function() {
+    assert.ok(utils.isValidEmail('foo@bar'));
+  });
+
+  it('should be invalid', function() {
```
kumar303: this test name is not unique

kumar303: could be `it('requires a TLD for email')`

kumar303:
r+wc

public/js/utils.jsx
muffinresearch: Ok cool - I'll look at cleaning this up in a follow-up.

PR #382, opened 2015-09-02
Show error on undefined action constants (fixes #229)

### LINKED ISSUE #229: Protect against undefined action types better
When creating redux stores it's easy to do something like this:
````
export myStore(state, action) {
  if (action.type === actionTypes.NEW_THING) {
    ...
  }
}
````
and forget to put `NEW_THING` in the constants. This will create a true expression! That's because `action.type` seems to be undefined sometimes for a weird reason (might be a bug).

I took a stab at it using getters but didn't get it working so I'm just filing this to try it again later.

### PR DESCRIPTION
**Pull request by muffinresearch**
action.type is undefined error console message

### DIFF
--- a/public/js/data-store.js
+++ b/public/js/data-store.js
@@ -4,6 +4,10 @@ import * as reducers from 'reducers';
 
 function logger({ getState }) {
   return next => action => {
+    if (typeof action.type === 'undefined') {
+      console.error('action.type is undefined.',
+        'Check that the action is defined in constants/action-types.js');
+    }
     console.info('redux: dispatching', action);
     const result = next(action);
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: public/js/data-store.js [hunk 1]
```diff
@@ -4,6 +4,10 @@ import * as reducers from 'reducers';
 
 function logger({ getState }) {
   return next => action => {
+    if (typeof action.type === 'undefined') {
+      console.error('action.type is undefined.',
+        'Check the action is defined in constants/action-types.js');
```
kumar303: nit: 'Check that the ...'

kumar303:
Nice!

r+wc

PR #384, opened 2015-09-02
Convert pay method to use a pay-method list (fixes #383, fixes #323)

### LINKED ISSUE #323: Change 'pay-methods' view to use a pay method choice.
Note this probably provides a good chance to rename card-choice to pay-method choice (connects to #310).

### LINKED ISSUE #383: Convert pay method buttons to be links as per UX
To avoid confusion the links that are styled as buttons should become links.

muffinresearch:
Before:

Before pay methods screenshot

After: 

After pay methods screenshot


kumar303:
r+

PR #386, opened 2015-09-02
Store/retrieve app state all in one place (fixes #347)

### LINKED ISSUE #347: Consolidate application error state
Using two different reducers, one at `reducers/app` and one at `reducers/management`, we store application state errors in both places (grep for `actionTypes.APP_ERROR`). We only need to store that in one place, probably just in `reducers/app`. The management app can access `state.app` to get the error.

kumar303:
Fixes https://github.com/mozilla/payments-ui/issues/347

### DIFF
--- a/public/js/reducers/management.js
+++ b/public/js/reducers/management.js
@@ -2,7 +2,6 @@ import * as actionTypes from 'constants/action-types';
 
 
 export const initialMgmtState = {
-  error: null,
   tab: null,
   view: null,
   viewData: {},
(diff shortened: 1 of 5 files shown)


public/js/reducers/management.js
```diff
@@ -21,12 +20,6 @@ export default function management(state, action) {
         });
       }
       return state;
-    case actionTypes.APP_ERROR:
-      return Object.assign({}, initialMgmtState, {
-        error: {
-          debugMessage: action.error.debugMessage,
-        },
-      });
```
kumar303: huh, I guess this wasn't tested

muffinresearch:
This is going to clash with the redux 2.0 upgrade patch. :/

kumar303:
I'll fix this up after https://github.com/mozilla/payments-ui/pull/392

muffinresearch:
r+wpt

Issue #353, opened 2015-08-20
Make tests not depend on payments-config


kumar303:
Many tests need to use products that have known properties such as a recurring monthly subscription or donation, etc. For convenience we've been hard coding product IDs in the tests for products like `mozilla-concrete-brick` and `mozilla-foundation-donation`. We should refactor these tests to use a product defined from scratch having the properties we want.

PR #387, opened 2015-09-02
Decide to sign in w/ user identification config (fixes #372)

### LINKED ISSUE #372: Use the payments-config product data to control how the user indentification is handled.
We are currently look at `product.seller.kind === 'donation'` we should use product.user_indentification for that instead.

kumar303:
Fixes https://github.com/mozilla/payments-ui/issues/372

### DIFF
--- a/public/js/views/transaction/product-pay.jsx
+++ b/public/js/views/transaction/product-pay.jsx
@@ -11,7 +11,6 @@ import { gettext } from 'utils';
 export default class ProductPay extends Component {
 
   static propTypes = {
-    // Amount to pay, which only applies to things like donations.
     braintreeToken: PropTypes.string.isRequired,
     cardSubmissionErrors: PropTypes.object,
     processPayment: PropTypes.func.isRequired,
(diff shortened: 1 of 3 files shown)


kumar303:
I can't really write a better test for this until we have https://github.com/mozilla/payments-ui/issues/353 (the sign-in logic is currently tested though)

public/js/views/transaction/product-pay.jsx
```diff
@@ -11,7 +11,6 @@ import { gettext } from 'utils';
 export default class ProductPay extends Component {
 
   static propTypes = {
-    // Amount to pay, which only applies to things like donations.
```
muffinresearch: Thanks for cleaning this up.

muffinresearch:
r+wc

kumar303:
> Was there other places the donations condition was checked. 

There are but I think they make sense as is because they are actually about donations and not about user identification:

https://github.com/mozilla/payments-ui/blob/master/public/js/views/transaction/complete-payment.jsx#L70
https://github.com/mozilla/payments-ui/blob/master/public/js/views/transaction/product-pay.jsx#L45
https://github.com/mozilla/payments-ui/blob/master/public/js/views/transaction/product-pay-chooser.jsx#L35

PR #394, opened 2015-09-03
Fix lint on tests (fixes #393)

### LINKED ISSUE #393: Lint tests
Seems tests aren't being lint-checked or the config is wrong.

muffinresearch:
Fixes #393

### DIFF
--- a/tasks/eslint.js
+++ b/tasks/eslint.js
@@ -3,7 +3,7 @@ module.exports = {
     src: [
       'public/js/**/*.js*',
       '!public/js/bundle.js',
-      'tests/*.*',
+      'tests/**/*.js*',
       'Gruntfile.js',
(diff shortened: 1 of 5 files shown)


tasks/eslint.js
```diff
@@ -3,7 +3,7 @@ module.exports = {
     src: [
       'public/js/**/*.js*',
       '!public/js/bundle.js',
-      'tests/*.*',
+      'tests/**/*.js*',
```
kumar303: d'oh

kumar303:
Nice!

r+ute (unless travis explodes)

tasks/eslint.js
muffinresearch: Yeah, I missed this when changing to have tests under specific dirs.

Issue #282, opened 2015-08-06
Use notifications for error messages


andymckay-limited-access:
In the management interface:

![7alhwojkqj7dr8eoz_hdsv3pmglkd4ccyrjzshxrice](https://cloud.githubusercontent.com/assets/1514/9134441/10c8751e-3cfc-11e5-9c84-88e6143f0471.png)

muffinresearch:
This just some leftover from  old layout.

muffinresearch:
Could use some ux-input here on how we want to represent general errors in the managment interface.

brampitoyo:
**Decision:** first, design notification flag, then we can use it to display errors and take action (e.g. “try again”) a la Gmail.

brampitoyo:
How about an error message like this one?

![ux-management-error](https://cloud.githubusercontent.com/assets/51007/9190637/83891cbe-404b-11e5-9988-aff86ea99a2f.png)

tofumatt:
As a notification that looks nice, but for errors in particular would some red colour be alright? Might draw a bit more attention and also signal something definitely went wrong.



- tofumatt

brampitoyo:
Yep. Some red would definitely be alright. How does this look?

![ux-management-wireframe-error](https://cloud.githubusercontent.com/assets/51007/9266733/4ab38fbe-4299-11e5-86ee-3142edbff9c4.png)

tofumatt:
Reads much more like an error 😃. Nice! 👍



- tofumatt

muffinresearch:
Relates to #413

PR #395, opened 2015-09-03
Show an error for invalid products (fixes #99)

### LINKED ISSUE #99: Show user friendly error when product ID is wrong
It's possible that the 3rd party who requests payment for a product will send us the wrong product ID. This could be due to a typo or an out of date configuration. In https://github.com/mozilla/payments-ui/pull/94/ we will throw a JS exception for this but that will result in a spinner. Instead, we should show the user an explanation of what happened. Something like "This product cannot be purchased due to a configuration error."

kumar303:
Fixes https://github.com/mozilla/payments-ui/issues/99

screenshot 2015-09-03 17 13 35

### DIFF
--- a/public/js/actions/app.js
+++ b/public/js/actions/app.js
@@ -1,9 +1,12 @@
 import * as actionTypes from 'constants/action-types';
 
 
-export function error(debugMessage) {
+export function error(debugMessage, {userMessage} = {}) {
   return {
     type: actionTypes.APP_ERROR,
-    error: {debugMessage: debugMessage},
+    error: {
+      debugMessage: debugMessage,
+      userMessage: userMessage,
+    },
   };
 }
--- a/public/js/apps/transaction/app.jsx
+++ b/public/js/apps/transaction/app.jsx
@@ -6,13 +6,14 @@ import { connect } from 'react-redux';
 import { bindActionCreators } from 'redux';
 
 import ErrorMessage from 'components/error';
+import Spinner from 'components/spinner';
 import DefaultSignIn from 'views/shared/sign-in';
 import DefaultTransaction from 'views/transaction';
 
 import * as appActions from 'actions/app';
 import * as userActions from 'actions/user';
 import * as products from 'products';
-import { parseQuery } from 'utils';
+import { gettext, parseQuery } from 'utils';
 
 
 export class TransactionApp extends Component {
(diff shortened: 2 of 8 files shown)


kumar303:
@brampitoyo does this look ok? I couldn't find a UX mock in the docs for this kind of stop-the-world error (I vaguely recall you spec'd this out though). Note that this is different than https://github.com/mozilla/payments-ui/issues/282 because it's for the purchase flow, not the management flow.

public/js/apps/transaction/app.jsx
```diff
@@ -34,55 +35,75 @@ export class TransactionApp extends Component {
 
   constructor(props) {
     super(props);
+    this.boundAppActions = bindActionCreators(appActions, this.props.dispatch);
```
muffinresearch: Can you use `props` instead of `this.props` here as they're passed?

tests/reducers/test.app-reducer.js
```diff
@@ -10,6 +10,7 @@ describe('App Reducer', function() {
     return {
       error: {
         debugMessage: 'some informative message',
+        userMessage: undefined,
```
muffinresearch: Is this for readability? Otherwise the unset key will be undefined by default.

public/js/apps/transaction/app.jsx
```diff
@@ -34,55 +35,75 @@ export class TransactionApp extends Component {
...
     } else {
-      console.log('rendering purchase flow');
-      return (
-        <Transaction
-          productId={state.productId}
-          userDefinedAmount={state.userDefinedAmount}
-        />
-      );
+
+      var SignIn = this.props.SignIn;
```
muffinresearch: We could probably move these into the de-structuring assignment above.

muffinresearch:
This looks fine to me. iirc this was as specced for the transaction flow in the past.

r+wc

public/js/actions/app.js
```diff
@@ -1,9 +1,10 @@
 import * as actionTypes from 'constants/action-types';
 
 
-export function error(debugMessage) {
+export function error(debugMessage, {userMessage} = {}) {
   return {
     type: actionTypes.APP_ERROR,
-    error: {debugMessage: debugMessage},
+    error: {debugMessage: debugMessage,
+            userMessage: userMessage},
```
muffinresearch: This might look nicer like so: 

```
error: {
  debugMessage: debugMessage,
  userMessage: userMessage,
},
```

public/js/apps/transaction/app.jsx
kumar303: oh yeah, heh

tests/reducers/test.app-reducer.js
kumar303: I was scratching my head on this but `assert.deepEqual` was failing without it. Maybe it looks at object keys first but it's curious how the key is available when its value is undefined.

muffinresearch: Oh strange. Maybe add a comment so it doesn't get factored out?

kumar303: the tests will fail :)

muffinresearch: mal-uh-whut

PR #396, opened 2015-09-04
Add already subscribed message (fixes #92)

### LINKED ISSUE #92: Cope with repeated subscription attempt
STR:
* leave ALLOW_REPEAT_PAYMENTS to False
* try a repeated payment
* payment-ui is stuck in a spinner

Payments-service throws an error:

```
payments_service.braintree.views:INFO buyer 1 is already subscribed to product 3 :/srv/payments-service/payments_service/braintree/views.py:80
[09/Jun/2015 18:23:10]"POST /api/braintree/subscriptions/ HTTP/1.0" 400 111
```

muffinresearch:
Fixes #92

* Errors are now overlaid
* Shows error on "already subscribed"

mozilla_payments

### DIFF
--- a/public/js/actions/subscriptions.js
+++ b/public/js/actions/subscriptions.js
@@ -4,6 +4,8 @@ import * as api from './api';
 import * as appActions from './app';
 import * as transactionActions from './transaction';
 
+import { gettext, arrayHasSubString } from 'utils';
+
 
 export function getUserSubscriptions(fetch=api.fetch) {
   return (dispatch, getState) => {
@@ -98,11 +100,19 @@ export function _createSubscription({dispatch, productId,
     console.log('Successfully subscribed + completed payment');
     dispatch(transactionActions.complete({userEmail: email}));
   }).fail($xhr => {
-    if (data.pay_method_nonce) {
-      dispatch({
-        type: actionTypes.CREDIT_CARD_SUBMISSION_ERRORS,
-        apiErrorResult: $xhr.responseJSON,
-      });
+    if ($xhr.status === 400 && $xhr.responseJSON &&
+        $xhr.responseJSON.error_response) {
+      var allErrors = $xhr.responseJSON.error_response.__all__;
+      if (allErrors && arrayHasSubString(allErrors, 'already subscribed')) {
+        dispatch(appActions.error('Subscription creation failed', {
+          userMessage: gettext('User is already subscribed to this product'),
+        }));
+      } else if (data.pay_method_nonce) {
+        dispatch({
+          type: actionTypes.CREDIT_CARD_SUBMISSION_ERRORS,
+          apiErrorResult: $xhr.responseJSON,
+        });
+      }
     } else {
       dispatch(appActions.error('Subscription creation failed'));
     }
--- a/public/js/apps/transaction/app.jsx
+++ b/public/js/apps/transaction/app.jsx
@@ -58,14 +58,20 @@ export class TransactionApp extends Component {
     };
   }
 
-  renderChild() {
-    var { app, user, SignIn, Transaction } = this.props;
-    var state = this.state;
-
+  renderError() {
+    var { app } = this.props;
     if (app.error) {
       console.log('rendering app error');
       return <ErrorMessage error={app.error} />;
-    } else if (!state.isValid) {
+    }
+    return null;
+  }
+
+  renderChild() {
+    var { user, SignIn, Transaction } = this.props;
+    var state = this.state;
+
+    if (!state.isValid) {
       // This renders a temporary loading state while we wait for
       // redux to re-render the component with an error to display.
       return <Spinner />;
@@ -106,6 +112,7 @@ export class TransactionApp extends Component {
   render() {
     return (
       <main>
+        {this.renderError()}
         {this.renderChild()}
       </main>
     );
--- a/public/js/utils.jsx
+++ b/public/js/utils.jsx
@@ -158,3 +158,13 @@ export function validateEmailAsYouType(email) {
     };
   }
 }
+
+export function arrayHasSubString(list, searchSubString) {
+  for (var i = 0; i < list.length; i += 1) {
+    var item = list[i] || '';
+    if (item.indexOf(searchSubString) > -1) {
+      return true;
+    }
+  }
+  return false;
+}
--- a/public/scss/_app-error.scss
+++ b/public/scss/_app-error.scss
@@ -1,12 +1,13 @@
 .app-error {
-    background: rgba(255,255,255,0.8);
+    background: rgba(255,255,255,0.9);
     bottom: 0;
     height: 100%;
     left: 0;
+    padding: 10px;
     position: absolute;
     right: 0;
     top: 0;
-    z-index: 10;
+    z-index: 30;
 
     .msg {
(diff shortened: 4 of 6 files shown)


public/js/actions/subscriptions.js
```diff
@@ -98,11 +100,20 @@ export function _createSubscription({dispatch, productId,
     console.log('Successfully subscribed + completed payment');
     dispatch(transactionActions.complete({userEmail: email}));
   }).fail($xhr => {
-    if (data.pay_method_nonce) {
-      dispatch({
-        type: actionTypes.CREDIT_CARD_SUBMISSION_ERRORS,
-        apiErrorResult: $xhr.responseJSON,
-      });
+    console.log($xhr);
```
kumar303: Is this still needed? I've actually been wondering if we should log requests more verbosely (in the `fetch()` wrapper) because it's really annoying when you hit a bug and then realize you forgot to turn on XHR logging in the devtools.

public/js/actions/subscriptions.js
```diff
@@ -98,11 +100,20 @@ export function _createSubscription({dispatch, productId,
...
+    if ($xhr.status === 400 && $xhr.responseJSON &&
+        $xhr.responseJSON.error_response) {
+      var allErrors = $xhr.responseJSON.error_response.__all__;
+      if (allErrors && hasErrorSubString(allErrors, 'already subscribed')) {
```
kumar303: this is probably ok for now, but it makes a really good case for implementing https://github.com/mozilla/payments-service/issues/50

public/js/actions/subscriptions.js
```diff
@@ -98,11 +100,20 @@ export function _createSubscription({dispatch, productId,
...
+        dispatch(appActions.error('Subscription creation failed', {
+          userMessage: gettext('User is already subscribed to this product'),
+        }));
+      } else if (data.pay_method_nonce) {
+        dispatch({
+          type: actionTypes.CREDIT_CARD_SUBMISSION_ERRORS,
+          apiErrorResult: $xhr.responseJSON,
```
kumar303: I think it would make a lot more sense if dispatching this error was no different than dispatching the 'User is already subscribed...' error. Did you look into that? Is it too tricky to merge the two? They seem like they are trying to accomplish the same thing. Maybe we can just get rid of `CREDIT_CARD_SUBMISSION_ERRORS` altogether and use `appActions.error()` ?

public/js/apps/transaction/app.jsx
```diff
@@ -58,14 +58,20 @@ export class TransactionApp extends Component {
...
+    return null;
+  }
+
+  renderChild() {
+    var { user, SignIn, Transaction } = this.props;
+    var state = this.state;
+
+    if (!state.isValid) {
       // This renders a temporary loading state while we wait for
       // redux to re-render the component with an error to display.
```
kumar303: I guess this comment might be a little misleading now. Perhaps it should just explain how since the component is an error state, nothing else can be rendered.

public/js/utils.jsx
```diff
@@ -158,3 +158,12 @@ export function validateEmailAsYouType(email) {
     };
   }
 }
+
+export function hasErrorSubString(errorList, searchSubString) {
+  for (var i = 0; i < errorList.length; i += 1) {
```
kumar303: `i++` for hacker street cred? jk

public/js/utils.jsx
```diff
@@ -158,3 +158,12 @@ export function validateEmailAsYouType(email) {
...
+    if (errorList[i].indexOf(searchSubString) > -1) {
```
kumar303: for paranoia, this should probably be `(errorList[i] || '').indexOf(...)`

public/js/utils.jsx
```diff
@@ -158,3 +158,12 @@ export function validateEmailAsYouType(email) {
     };
   }
 }
+
+export function hasErrorSubString(errorList, searchSubString) {
```
kumar303: you could just call this `arrayHasSubString()` because it seems like it could be used for any array.

public/scss/_app-error.scss
```diff
@@ -1,12 +1,13 @@
...
+    background: rgba(255,255,255,0.9);
     bottom: 0;
     height: 100%;
     left: 0;
+    padding: 10px;
     position: absolute;
     right: 0;
     top: 0;
-    z-index: 10;
+    z-index: 30;
```
kumar303: something for the future (maybe): http://www.sitepoint.com/better-solution-managing-z-index-sass/ This seems like a better way to do handle z-index.

tests/actions/test.subs-actions.js
```diff
@@ -134,6 +137,36 @@ describe('subscription actions', function() {
...
+
+    it('should dispatch a generic error', function() {
+      var fetch = helpers.fakeFetch({
+        result: 'fail',
+      });
+      createSubscription({fetch: fetch});
+      var action = dispatchSpy.firstCall.args[0];
+      assert.equal(action.type, actionTypes.APP_ERROR);
+      assert.include(action.error.debugMessage, 'Subscription creation failed');
+    });
```
kumar303: nice tests

public/js/actions/subscriptions.js
```diff
@@ -98,11 +100,20 @@ export function _createSubscription({dispatch, productId,
...
     dispatch(transactionActions.complete({userEmail: email}));
   }).fail($xhr => {
-    if (data.pay_method_nonce) {
-      dispatch({
-        type: actionTypes.CREDIT_CARD_SUBMISSION_ERRORS,
-        apiErrorResult: $xhr.responseJSON,
-      });
+    console.log($xhr);
+    if ($xhr.status === 400 && $xhr.responseJSON &&
+        $xhr.responseJSON.error_response) {
```
kumar303: I bet we'll need this again for another action so we should make a mental note to wrap this up somehow (later).

kumar303:
it would be cool to merge subscription error dispatches if not too much work, otherwise r+wc

public/js/actions/subscriptions.js
muffinresearch: Yeah I didn't want to block on that. But definitely I thought the same doing this.

public/js/actions/subscriptions.js
muffinresearch: The difference is that card submission errors are much lower level. Whilst it's the same data it doesn't result in a hard failure. It's things like field errors.

public/js/utils.jsx
muffinresearch: Hmm, lint is set to not like i++; We should probably change that but that's why.

public/scss/_app-error.scss
muffinresearch: Yeah, the thing I need to trump was 10 so it's not 99999999 :). But yeah we could deffo consider using something fancy if it gets to be an issue. Probably just vars would do it.

public/js/utils.jsx
muffinresearch: Good idea!

public/js/actions/subscriptions.js
muffinresearch: Yeah that might be nice. Even if it's just status and responseJSON / responseText etc.

Issue #411, opened 2015-09-08
Show general payment processing errors


kumar303:
When processing a payment, the user does not see any general processing error, i.e. one not associated with a form field. They see either an empty credit card form or a spinner but no error message. This is pretty bad because there are a lot of errors that could happen like: https://developers.braintreepayments.com/javascript+python/reference/general/testing#unsuccessful-credit-card-verification

There is a UX mock for how this should look (the last screen here):
![ux-pay-form-error](https://cloud.githubusercontent.com/assets/55398/9740833/69899b86-561c-11e5-8c9a-5e933742233f.png)

muffinresearch:
Relates to #413

muffinresearch:
Possibly crosses over with #95.

PR #414, opened 2015-09-08
Add notification action/reducer (fixes #413)

### LINKED ISSUE #413: Build notifications action/reducer
This is so we can use notifications for errors and information purposes.

muffinresearch:


kumar303:
might need a polyfill for `Array.from`

public/js/actions/notifications.js
```diff
@@ -0,0 +1,32 @@
+import * as actionTypes from 'constants/action-types';
+
+
+export function addNotification(text, { type='info', userDismissable=false,
+                                 autoHide=true } = {}) {
```
kumar303: indent is off by one

public/js/apps/management/app.jsx
```diff
@@ -48,12 +51,13 @@ export default class ManagementApp extends Component {
   }
 
   render() {
-
     const { app, dispatch, management, user } = this.props;
 
     console.log('rendering management app at tab:', management.tab);
     var qs = parseQuery(this.props.window.location.href);
     var accessToken = qs.access_token;
+    var boundNotificationActions = bindActionCreators(notificationActions,
```
kumar303: `boundNotifyActions` ?

public/js/apps/management/app.jsx
```diff
@@ -48,12 +51,13 @@ export default class ManagementApp extends Component {
...
+                                                     dispatch);
```
kumar303: indent is off by one

public/js/components/notification-list.jsx
```diff
@@ -0,0 +1,42 @@
...
+export default class NotificationList extends Component {
+
+  static propTypes = {
+    TransitionGroup: PropTypes.node,
+    notifications: PropTypes.array,
+    removeNotification: PropTypes.func.isRequired,
+  }
+
+  static defaultProps = {
+    TransitionGroup: ReactCSSTransitionGroup,
```
kumar303: I think `notifications` should be set to an empty array by default here, right? Or maybe it should be a required prop?

public/js/components/notification-list.jsx
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  render() {
+    var CSSTransitionGroup = this.props.TransitionGroup;
+    var items = this.props.notifications.map((item) => {
+      // this.props.notifications is an array of 2 item arrays.
+      var key = item[0];
+      var val = item[1];
```
kumar303: `var [key, val] = item` for the hotness!! :fire: :sunglasses:

public/js/components/notification-list.jsx
```diff
@@ -0,0 +1,42 @@
...
+      return (
+        <Notification
+          key={key}
+          handleAutoHide={this.props.removeNotification.bind(this, key)}
```
kumar303: is this better than `{() => this.props.removeNotification(key)}` ?

public/js/components/notification.jsx
```diff
@@ -0,0 +1,46 @@
...
+
+import cx from 'classnames';
+
+
+export default class Notification extends Component {
+
+  static propTypes = {
+    autoHide: PropTypes.bool,
+    handleAutoHide: React.PropTypes.func,
+    handleDismissClick: React.PropTypes.func,
```
kumar303: I think you could replace these two props with `removeItem` and just call it like `this.removeItem()` for each case.

public/js/components/notification-list.jsx
kumar303: ah, yes it is, because it needs to take extra arguments :) Carry on.

public/js/components/notification.jsx
kumar303: In looking at this more, I don't think you need to bind it either. You could just call `this.props.removeNotification(this.props.key)`.

public/js/components/notification.jsx
```diff
@@ -0,0 +1,46 @@
...
+    autoHide: true,
+    type: 'info',
+    userDismissable: false,
+  }
+
+  componentDidMount() {
+    if (this.props.userDismissable === false &&
+        this.props.autoHide === true) {
+      this.props.handleAutoHide({delay: 5000});
+    }
```
kumar303: is there a realistic else case? i.e. notices that can't be dismissed and never go away? I can't think of one. You can probably ditch `autoHide` and make that the fallback behavior when the notice is not user dismissable.

kumar303: actually, I'll probably want a non-dimissable notice that doesn't auto hide for https://github.com/mozilla/payments-ui/issues/411

public/js/reducers/app.js
```diff
@@ -17,6 +19,28 @@ export default function app(state, action) {
       return Object.assign({}, state, {
         csrfToken: action.csrfToken,
       });
+    case actionTypes.ADD_NOTIFICATION:
+      var newNotifications = state.notifications.slice(0);
+      var notificationMap = new Map(newNotifications);
+      var notification = notificationMap.set(getId(), action.data);
+      newNotifications = [...notificationMap];
```
kumar303: huh, it's surprising to me that Map preserves order when deconstructed to an array but the docs say indeed it does.

public/js/reducers/app.js
```diff
@@ -1,9 +1,11 @@
 import * as actionTypes from 'constants/action-types';
+import { getId } from 'utils';
 
 
 export const initialAppState = {
   csrfToken: null,
   error: null,
+  notifications: [],
```
kumar303: It would be nice to comment how this looks will look like `[[key, notice], [key, notice], ...]`

public/js/actions/notifications.js
```diff
@@ -0,0 +1,32 @@
...
+  return (dispatch) => {
```
kumar303: why does this need to be deferred? Won't dispatching an action with `{meta: {delay: ...}}` already defer it in the middleware?

tests/apps/test.management-app.jsx
```diff
@@ -60,9 +60,9 @@ describe('management/app', function() {
     var mgmt = TestUtils.findRenderedComponentWithType(
       view, FakeManagement
     );
-    // The first child should be the sign-in component.
+    // The second child should be the sign-in component.
     // Make sure it gets the access token.
-    var signIn = mgmt.props.children[0];
+    var signIn = mgmt.props.children[1];
```
kumar303: mmm, we should probably do this a better way, like, iterate the children and match on component type. Not for this patch though :)

tests/helpers.jsx
```diff
@@ -143,10 +143,11 @@ export function fakeFetch({xhrError={}, returnedData={},
 }
 
 
-export function stubComponent() {
+export function stubComponent(mockTagName) {
```
kumar303: `mockTagName='div'` for hotness

tests/reducers/test.app-reducer.js
```diff
@@ -52,4 +52,28 @@ describe('App Reducer', function() {
...
+      type: actionTypes.ADD_NOTIFICATION,
+      data: {text: 'foo', type: 'info'},
+    });
+    var id = app.notifications[0][0];
+    var app = appReducer(initialAppState, {
+      type: actionTypes.REMOVE_NOTIFICATION,
+      id: id,
+    });
+    assert.equal(app.notifications.length, 0);
+  });
```
kumar303: it would be nice to add a test that tries to remove a non-existant notice just to make sure there's no exception

tests/reducers/test.app-reducer.js
```diff
@@ -52,4 +52,28 @@ describe('App Reducer', function() {
...
+                     {text: 'foo', type: 'info'});
+  });
+
+  it('should remove a notification', function() {
+    var app = appReducer(initialAppState, {
+      type: actionTypes.ADD_NOTIFICATION,
+      data: {text: 'foo', type: 'info'},
+    });
+    var id = app.notifications[0][0];
+    var app = appReducer(initialAppState, {
```
kumar303: this test will always pass :) It should be: `var app = appReducer(app, ...`

tests/test.utils.jsx
```diff
@@ -186,3 +186,12 @@ describe('utils.arrayHasSubString', function() {
   });
 
 });
+
+
+describe('utils.getId', function() {
+
+  it('should produce a unique id', function() {
+    assert.notEqual(utils.getId(), utils.getId());
```
kumar303: :+1:

kumar303:
I had a couple open questions and I think you can ditch the deferred actions otherwise r+wc; let me know if you want another look.

public/js/actions/notifications.js
muffinresearch: I was trying to work around the timeout issues. But yes this is no longer necessary. Good catch.

tests/apps/test.management-app.jsx
muffinresearch: Yeah :+1:

muffinresearch:
@kumar303 I ended up ditching all our standalone shims and opted for suing the babel polyfills instead. The standalone `Array.from` shim didn't work (tests failed when it was used). I think this is a better bet anyway as it means we're using babel blessed polyfills alongside babel which is less likely to break compatibility.

muffinresearch:
Hopefully sauce labs will agree :)

public/js/components/notification-list.jsx
muffinresearch: I made it required.

PR #417, opened 2015-09-09
Use notification for all errors (fixes #282)

### LINKED ISSUE #282: Use notifications for error messages

muffinresearch:
fixes #282 

This branch:

* Kills errors in favour of error notifications across the board
* Introduces error codes mainly for better error testing.
* fixes up the styling for notifications in the transaction app.

mozilla_payments


public/js/actions/notifications.js
```diff
@@ -1,14 +1,58 @@
...
+
+export function addNotification({ text, errorCode,
+                                  type='info', userDismissable,
+                                  autoHide } = {}) {
+
+  if (type === 'error' && !text) {
+    text = gettext('Internal error. Please try again later.');
+  }
+
+  if (type === 'info' && !text) {
```
kumar303: I think this could just be `if (!text)...` because `text` is set up above for error types

public/js/actions/notifications.js
```diff
@@ -1,14 +1,58 @@
...
+
+  // Default to error type being userDismissable.
+  if (type === 'error' && typeof userDismissable === 'undefined') {
+    userDismissable = true;
+  }
+
+  // Default info type to being hidden on delay.
+  if (type === 'info' && typeof autoHide === 'undefined') {
+    autoHide = true;
+  }
```
kumar303: The logic for setting up defaults would be easier to follow if there were two distinct actions:
````
export function addError(text, {type='error', ...}) {
  return addNotification(...);
}

export function addInfo(text, {type='info', ...}) {
  return addNotification(...);
}
````

File: public/js/actions/notifications.js [hunk 3]
```diff
@@ -1,14 +1,58 @@
...
+
+  if (type === 'error' && (!errorCode ||
+                           typeof errorCodes[errorCode] === 'undefined')) {
+    console.error('errorCode', errorCode);
```
kumar303: :fried_shrimp:

public/js/actions/notifications.js
kumar303: duh, I see you have separate functions already. Maybe move all the `if (type==='error')` checks and so on to each respective function? You could just make `addNotification` private so that it's never used directly.

public/js/actions/notifications.js
```diff
@@ -1,14 +1,58 @@
...
+    throw new Error('An error notification requires an errorCode ' +
+                    'defined in constants/error-codes');
+  }
+
+  console.log('autoHide', autoHide);
+  console.log('userDismissable', userDismissable);
```
kumar303: :fried_shrimp: :cocktail:

public/js/actions/subscriptions.js
```diff
@@ -104,8 +107,10 @@ export function _createSubscription({dispatch, productId,
         $xhr.responseJSON.error_response) {
       var allErrors = $xhr.responseJSON.error_response.__all__;
       if (allErrors && arrayHasSubString(allErrors, 'already subscribed')) {
-        dispatch(appActions.error('Subscription creation failed', {
```
kumar303: the old approach would log the message 'subscription creation failed' which I think is still important

public/js/actions/notifications.js
kumar303: oh, was this not for debugging? If not, could you preface it with something like 'notifying about error with code...' ?

public/js/actions/notifications.js
kumar303: looks like debugging

public/js/actions/notifications.js
kumar303: I think it would be nice to log the error code here

andymckay:
:clap: thanks for adding that, pretty!

public/js/constants/error-codes.js
```diff
@@ -0,0 +1,17 @@
...
+export const ONE_TIME_PAYMENT_FAILED = 'ONE_TIME_PAYMENT_FAILED';
+export const PAY_METHOD_DELETION_FAILED = 'PAY_METHOD_DELETION_FAILED';
+export const PAY_METHOD_GET_FAILED = 'PAY_METHOD_GET_FAILED';
+export const PRODUCT_ID_INVALID = 'PRODUCT_ID_INVALID';
+export const SUBS_BY_PAY_METHOD_GET_FAILED = 'SUBS_BY_PAY_METHOD_GET_FAILED';
+export const SUBS_GET_FAILED = 'SUBS_GET_FAILED';
+export const SUB_CREATION_FAILED = 'SUB_CREATION_FAILED';
+export const SUB_UPDATE_FAILED = 'SUB_UPDATE_FAILED';
+export const TRANSACTIONS_GET_FAILED = 'TRANSACTIONS_GET_FAILED';
+export const UNSPECIFIED_ERROR = 'UNSPECIFIED_ERROR';
```
kumar303: I don't see this used anywhere. Do we need it? It would be nice to always specify the error.

tests/helpers.jsx
```diff
@@ -189,3 +189,23 @@ export class FakeBraintreeClientError {
...
+  assert.ok(typeof action.data.text === 'string',
+            'action.data.text should be a string');
+  assert.include(action.data.text, textSubString,
+                 'action.data.text should include substring');
+}
+
+export function assertNotificationErrorCode(action, errorCode) {
+  assertNotificationType(action, 'error');
+  assert.equal(action.data.errorCode, errorCode,
+               'action.data.errorCode must match errorCode');
```
kumar303: I don't think you need a third argument here. Without it, the message will log the actual and expected value which is more useful.

kumar303:
Nice! Looks like a pretty straight forward patch. I think `showError` and `showInfo` should have their own respective validation; the other comments are minor. r+wc

public/js/constants/error-codes.js
muffinresearch: Yeah, I can kill that now it throws. I thought it made sense to not have an error with no code.

public/js/actions/notifications.js
muffinresearch: Good call. That's a nice iteration.

tests/helpers.jsx
muffinresearch: It does both.

kumar303: no way! I didn't know it did both. That's awesome

PR #418, opened 2015-09-10
Update eslint config

muffinresearch:
* Enforces prop quoting only when necessary and enforces constistency
* Allow not having spaces between operators e.g: `i++`
* I've cleaned up all the unnecessary prop quotes.


tofumatt:
:+1: r+

PR #421, opened 2015-09-10
Add more robust processing indicators (fixes #411)

### LINKED ISSUE #411: Show general payment processing errors

kumar303:
This adds processing state and also fixes broken forms after handling errors. 

Fixes: https://github.com/mozilla/payments-ui/issues/411

The approach:
- adds a flexible mechanism to store/retrieve processing state for any component
- any action should dispatch begin/stop notifications when processing data
  - I only implemented this for the credit card submission and saved pay method submission
- a lot of components are now watching state with `connect()` but for simplicity I mostly tested the raw class objects, not provider-wrapped objects. I think this works ok because dispatch isn't really involved in the tests.

Here is a screenshot of an enabled submit button after viewing errors which was not possible before. What isn't shown here is that the spinner only occurs during submission, then it goes away.

screenshot 2015-09-10 18 52 47


public/scss/_app-error.scss
```diff
@@ -1,21 +0,0 @@
-.app-error {
```
kumar303: I don't think this is used anymore, right?

public/js/utils.jsx
```diff
@@ -169,6 +169,6 @@ export function arrayHasSubString(list, searchSubString) {
   return false;
 }
 
-export function getId() {
-  return '_' + Math.random().toString(36).substr(2, 9);
+export function getId({prefix=''} = {}) {
+  return prefix + '_' + Math.random().toString(36).substr(2, 9);
```
muffinresearch: Nice!

public/scss/_app-error.scss
muffinresearch: Yep it's good to go.

muffinresearch:
I suppose one thing about that specific example in the screenshot is that the user shouldn't be able to keep submitting as it will always fail. Maybe something to think about for the future is a different class of error notifications. E.g. showFatal which goes back to more of a modal error?

public/js/actions/processing.js
```diff
@@ -0,0 +1,22 @@
...
+
+
+export function stopProcessing(processingId) {
+  return processingAction(processingId, actionTypes.STOP_PROCESSING);
+}
+
+
+function processingAction(processingId, actionType) {
+  if (!processingId) {
+    throw new Error('processingId cannot be empty');
```
muffinresearch: :+1:

public/js/components/card-form.jsx
```diff
@@ -147,13 +150,12 @@ export default class CardForm extends Component {
...
     }
-    this.props.handleCardSubmit(formData, opts);
+    this.props.handleCardSubmit(formData, this.processingId, opts);
   }
 
   mapErrorsToFields(errors) {
     console.log('mapping submission errors to form fields', errors);
-    // TODO: map non-braintree errors like __all__
-    // TODO: map braintree errors for unexpected fields (such as `plan_id`)
+    // Note that __all__ errors will be rendered as notifications instead.
```
muffinresearch: Thx!

tests/views/test.product-pay.jsx
```diff
@@ -10,21 +12,33 @@ describe('ProductPay', function() {
...
+              cardSubmissionErrors={{}}
+              productId={productId}
+              processPayment={fakePaymentProcessor}
+            />
+          );
+        }}
+      </Provider>
+    );
+
+    return TestUtils.findRenderedComponentWithType(
```
muffinresearch: Maybe we should have added a helper to do by Type just for consistency with all the other stuff we have.

public/js/views/management/add-pay-method.jsx
```diff
@@ -33,7 +35,7 @@ export default class AddPayMethod extends Component {
         <CardForm
           submissionErrors={this.props.cardSubmissionErrors}
           submitPrompt={gettext('Add')}
-          handleCardSubmit={(card) => this.handleCardSubmit(card)}
+          handleCardSubmit={this.handleCardSubmit.bind(this)}
```
muffinresearch: Could you make the handleCardSubmit a fat arrow func then you probably don't need the bind.

e.g:
```
handleCardSubmit = (creditCard, processingId) => {

}
```

tests/reducers/test.processing-reducer.js
```diff
@@ -0,0 +1,36 @@
...
+                     initialProcessingState);
+  });
+
+  it('should store true when a component begins processing', function() {
+    var procId = 'some-id';
+    var state = processingReducer(initialProcessingState, {
+      type: actionTypes.BEGIN_PROCESSING,
+      processingId: procId,
+    });
+    assert.equal(state[procId], true);
```
muffinresearch: Could just used `assert.ok`.

muffinresearch:
Excellent stuff! r+wc

tests/views/test.product-pay.jsx
kumar303: I was thinking this Provider/find pattern could probably be wrapped up a little better too

public/js/views/management/add-pay-method.jsx
kumar303: one thing I like about bind is that you don't have to update it when the signature changes. In this case, it actually took me a while to track down that the lambda wasn't passing the right args :(

kumar303:
> the user shouldn't be able to keep submitting as it will always fail

In the CardForm case, it has no way to know that the form is invalid which is why the button is enabled. One thing we could do is put a `wasSuccessful` property on the processing state. That would be a pretty simple fix. Maybe I'll try it here and punt to another patch if it gets complicated.

kumar303:
> put a wasSuccessful property on the processing state

actually, nevermind. That's not the right place for it because processing state is ephemeral in nature.

PR #423, opened 2015-09-11
Add description to all grunt tasks

muffinresearch:
Thanks to @tofumatt I realise the description arg can be used for alias tasks too. (I've made a PR to fix the grunt-docs to mention this because they currently make it sounds like you can only use a description on function tasks.).

This updates all the tasks to have descriptions. I've removed comments where the description string covers the same thing as the comment.


tofumatt:
Hah, very nice! You beat me to this, I was thinking of adding these too 😀

R+

PR #424, opened 2015-09-11
Remove grunt-react

muffinresearch:


tofumatt:
It's deprecated so makes sense to me. R+

Issue #420, opened 2015-09-10
Can't enter a new credit card after first one is declined


kumar303:
To reproduce:
- Subscribe to a product using any [test number](https://developers.braintreepayments.com/javascript+python/reference/general/testing) to trigger a gateway error such as 'card was declined'
- See the declined error (screenshot)
- try to enter a new card number

You cannot enter a new number because the error hangs around and the Subscribe button is disabled.

![screenshot 2015-09-10 12 49 24](https://cloud.githubusercontent.com/assets/55398/9796223/a2ddf48c-57ba-11e5-889b-d92763a331d6.png)

muffinresearch:
So one issue with this is that the spinner is started on submit. Not sure how we can turn that off. My first thought was if all the XHR returns the deferred then we can do something with that. Sadly the tokenization is all callbacks so that makes it harder in this case. Actions might be possible but seems a bit too far detached from the form to use such a generalized solution.

kumar303:
yeah, I'm actually fixing the spinner in https://github.com/mozilla/payments-ui/issues/411 ; I have an idea for how to do it. Hopefully should be ready for r? when you read this.

muffinresearch:
Excellent, looking at that now.

PR #427, opened 2015-09-11
Show credit card entry errors in mgmt (fixes #426)

### LINKED ISSUE #426: add card error is not visible in management
If you try to add a card on the management screens and it causes an API error, you don't see the error even though the dispatch happens. Even though I've been working on code related to this I'm pretty sure it was always broken.

kumar303:
Fixes https://github.com/mozilla/payments-ui/issues/426

After the fix the screen looks like this:
screenshot 2015-09-11 15 04 52

However, you still can't proceed because of https://github.com/mozilla/payments-ui/issues/420


muffinresearch:
r+

PR #434, opened 2015-09-29
Updated react-redux to version 3.0.1

tofumatt:
:rocket:

One of your dependencies has just updated, so this PR bumps the corresponding version number in your `package.json`.

You can now check that the dependency update doesn't break your code, and then release the new version of your software safe in the knowledge that it will stay in this working state, regardless of _when_ your users do `$ npm install`.

Have a good day!

---

This pull request was created by [greenkeeper.io](http://greenkeeper.io/).
It keeps your software, up to date, all the time.

<sub>
Tired of seeing this sponsor message? Upgrade to the supporter plan!
You'll also get your pull requests faster :zap:
</sub>

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "react": "0.13.3",
     "react-json-table": "0.0.2",
     "react-maskedinput": "2.0.0",
-    "react-redux": "2.0.0",
+    "react-redux": "3.0.1",
     "redux": "2.0.0",
     "redux-thunk": "0.1.0"
   },


tofumatt:
Moving to a major new version seems like something we should manually test out, but all the tests pass so that's promising.

muffinresearch:
Yeah for redux upgrades it makes sense to do redux + react-redux together. The other one failed so that'll need looking into anyway.

muffinresearch:
v4.0 is coming and should be changed along with the update to react 0.14 See https://github.com/rackt/react-redux/releases/tag/v4.0.0

muffinresearch:
Going to close this as it's been superceded by a newer version.

PR #552, opened 2015-12-15
Delete node modules

muffinresearch:
To reduce the noise whilst this is shelved this removes the node_modules from the tree.


tofumatt:
Is this so Greenkeeper things fail less? I forget why these were checked in.

tofumatt:
If so: r+ wfm

muffinresearch:
> Is this so Greenkeeper things fail less? I forget why these were checked in.

Yeah it was checked in for reliability (and concreate reproducibility) but I'm trying to clear the backlog of the lib updates so getting rid of this for now makes things a bit easier.

Issue #463, opened 2015-10-18
Reduce noise in tests


tofumatt:
The test output is _really_ noisy right now, to the point that Travis CI won't show all test output online and requires you to download the raw test output. We should prevent all the log messages and probably the `cjs` stuff from outputting, at least unless something is wrong.

PR #568, opened 2015-12-26
silence logging from components while running tests

leethomas:
Hey guys, this addresses https://github.com/mozilla/payments-ui/issues/463.

I tried using mocha & karma config options to silence all of the logging coming from the client code, but nothing I tried worked. I finally decided to just redefine console.log while tests are running. The test output is a lot cleaner now.

tests


tests/test-loader.js
```diff
@@ -3,5 +3,11 @@
 // See: https://github.com/webpack/karma-webpack#alternative-usage
 import 'babel-core/polyfill';
 
+// silence logging in components
+var logLevels = ['log', 'warn', 'info', 'error'];
```
tofumatt: What about `debug`? It obviously doesn't get used but adding it would be future-proof.

karma.shared.js
```diff
@@ -44,5 +44,6 @@ module.exports = {
   webpack: newWebpackConfig,
   webpackServer: {
     noInfo: true,
+    quiet: true
```
tofumatt: This is a super-nitpick but adding a trailing comma here would keep it consistent (though it's mostly our fault for not having style tests in the repo).

tofumatt:
That is _sooooo_ much easier to look at, thank you! :+1: 

I have a question about `debug` level console stubbing, but otherwise r+ from me.

Let me know what you think--feel free to either add `debug` and the comma or just let me know and I can amend the commit before merging myself.

leethomas:
Adding debug is a good idea. I'll also add the trailing comma since consistency is major important, even if this is small. Happy to help out!

tofumatt:
Thanks for the quick reply :smile: 

I combined the commits into one just to make things nice and atomic. Merged in 7f210dd1497b5a6da449143bdc7b0c9d10969c43.

Cheers!

PR #617, opened 2016-01-27
chore(package): update grunt-i18n-abide to version 0.0.18

muffinresearch:
Update to get branch run against master.

Once travis has run a build re-running the tests only runs the tests against the build made at the time the pr was made. As a result changes to master since that point in time are not reflected in the re-run build.

Since the new version of karma should fix all the socket related forEach bugs we want to run this change against master.


muffinresearch:
Hmm npm seems to not have this package for some reason - it's a 404. Have asked for support.

